PR #20, opened 2017-10-06
Updated authorization component to use drop down edit behavior.

jimmyca15:


jimmyca15:
This never got checked in with the last update of the UI.

PR #32, opened 2017-12-04
Enhance/monitoring

jimmyca15:
The monitoring graphs were showing average values such as average requests per second. This change updates the monitoring view to show current values rather than calculated averages.


jimmyca15:
Changes reviewed with @drago-draganov

PR #33, opened 2017-12-04
Navigation Bar Resize + Breadcrumbs

jimmyca15:
Bread crumbs have been introduced to provide a more familiar navigation mechanism to users. Previously we used a vertical breadcrumb system that was not intuitive to many first time users.

While we added breadcrumbs we took the time to shrink the navigation bar which took up more space than necessary. This improves the amount of useful content that we can fit on the screen.


jimmyca15:

PR #34, opened 2017-12-04
Unify application pool details in web site creation dialog

jimmyca15:
Addresses #24


jimmyca15:

PR #35, opened 2017-12-04
Enhanced contrast of the application.

jimmyca15:
A review of the contrasts between font colors and primary/background colors led to a need to increase contrast. This PR introduces a minimum contrast of 4.5 for ease of reading and prolonged usage.


jimmyca15:

PR #37, opened 2017-12-04
Root folder edit uses dropdown + fix inconsistencies

jimmyca15:
* The process of editing and creating root folders now uses the drop down edit dialog.

* Certain behaviors such as clicking out of a new folder dialog were inconsistent with modern file explorer design.

* Removed unnecessary buttons from folder selector dialog such as when specifying a web site's physical path.


jimmyca15:

PR #41, opened 2017-12-06
Added capability to specify a favicon.

jimmyca15:


jimmyca15:
@shirhatti 

I took the approach that was used for allowing ga to be specified.

shirhatti:
ga?

PR #43, opened 2018-02-26
Update Angular to V5 and Use VS 2017 to build

jimmyca15:
The update required a small amount of changes to code.


drago-draganov (approved):
Looks good!

PR #207, opened 2018-09-27
Move to angular cli and create WAC entrypoint

shhsu:
I will remove the visual studio files in a separate PR


jimmyca15:
Since this PR is so big do you plan on writing a description for all the changes, especially the architectural ones, contained in this?

src/Microsoft.IIS.WebManager/angular-cli.json
```diff
@@ -0,0 +1,49 @@
+{
```
shhsu, line 1: remove visual studio project files in favor of angular-cli.json since this is a more widely used. My next step would be to remove the self hosting projects since we are not going that route anymore.

src/Microsoft.IIS.WebManager/app/common/component-loader.ts
```diff
@@ -1,11 +1,155 @@
...
-        return System.import(path)
-            .then(m => {
-                return m[name]
-            });
+export class ComponentLoader {
+    public static async LoadAsync(compiler: Compiler, component: ComponentReference): Promise<ModuleWithComponentFactories<{}>> {
+        // cannot assign module type as property of ComponentReference to avoid circular reference
+        var moduleType : Promise<Type<{}>> = null
+        switch (component.component_name) {
+            case WebSiteListComponentName:
```
shhsu: Angular CLI statically analyze which code do bring in during compile time. I cannot pass a expression to System.Import function anymore. The parameter needs to be a string literal for the code to work

src/Microsoft.IIS.WebManager/app/common/component-loader.ts
```diff
@@ -1,11 +1,155 @@
...
+                moduleType = import('../webserver/websites/websites.module').then(m => {
```
shhsu, line 38: Angular CLI statically analyze which code do bring in during compile time. I cannot pass a expression to System.Import function anymore. The parameter needs to be a string literal for the code to work

src/Microsoft.IIS.WebManager/app/common/dynamic.component.ts
```diff
@@ -89,7 +83,7 @@ export class DynamicComponent implements OnInit {
         }
 
         let vRef = this.vcRef;
-        return DynamicComponent.CreateModuleWithComponentFactories(this.compiler, this._moduleName, this._modulePath)
+        return ComponentLoader.LoadAsync(this.compiler, this.module)
```
shhsu, line 86: Angular CLI statically analyze which code do bring in during compile time. I cannot pass a expression to System.Import function anymore. The parameter needs to be a string literal for the code to work

src/Microsoft.IIS.WebManager/app/common/selectable.ts
```diff
@@ -3,7 +3,7 @@ import { CommonModule } from '@angular/common';
 import { FormsModule } from '@angular/forms';
 
 @Injectable()
-class SelectableService {
+export class SelectableService {
```
shhsu, line 6: This export is required for angular cli compilation

src/Microsoft.IIS.WebManager/app/connect/connect.service.ts
```diff
@@ -19,7 +19,7 @@ export class ConnectService {
 
     private _store: ConnectionStore;
     private _client: HttpConnection;
-    private _pingTimeoutId: number;
+    private _pingTimeout: NodeJS.Timer;
```
shhsu, line 22: setTimeout/clearTimeout now takes an object

src/Microsoft.IIS.WebManager/app/files/file-editor.ts
```diff
@@ -94,9 +94,10 @@ export class FileEditor {
         this._dirty = false;
 
         this.getFileText().then(txt => {
-            this._text.next(txt);
-            this._dirty = false;
-           
+            if (txt) {
```
shhsu: es6 more rigorous type checking, txt might be null

src/Microsoft.IIS.WebManager/app/files/file-nav.service.ts
```diff
@@ -16,6 +14,9 @@ import { ApiFile, ApiFileType, ChangeType } from './file';
 import { Navigator } from './navigator';
 import { FilesService } from './files.service';
 
+import 'rxjs/add/operator/pairwise'
```
shhsu, line 17: This change took me an whole afternoon to figure out
Static methods import need to be placed after the class it refer to is imported. Otherwise, there's no compile time error but at run time these methods would just go missing.

src/Microsoft.IIS.WebManager/app/main/app-routing.module.ts
```diff
@@ -0,0 +1,37 @@
+import { NgModule } from '@angular/core';
```
shhsu, line 1: our old way of declaring routing no longer works. This is the newer standard

src/Microsoft.IIS.WebManager/app/main/app.wac.module.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        { provide: "WebServerService", useClass: WebServerService },
+        { provide: "WebSitesService", useClass: WebSitesService },
+        { provide: "AppPoolsService", useClass: AppPoolsService },
+        { provide: "FilesService", useClass: FilesService },
+        { provide: ErrorHandler, useClass: AppErrorHandler },
+        { provide: "Runtime", useClass: WACRuntime}
+    ]
+})
+export class WACAppModule {
```
shhsu, line 28: WAC app modules that injects more WAC specific runtime objects

src/Microsoft.IIS.WebManager/app/main/home.component.ts
```diff
@@ -24,14 +24,14 @@ import { OptionsService } from './options.service';
             <div class="sidebar" [class.nav]="_options.active">
                 <vtabs [markLocation]="true" (activate)="_options.refresh()">
                     <item [name]="'Web Sites'" [ico]="'fa fa-globe'">
-                        <dynamic [name]="'WebSiteListComponent'" [module]="'app/webserver/websites/websites.module#WebSitesModule'"></dynamic>
+                        <website-list-component></website-list-component>
```
shhsu: These objects seem to not need to be dynamic since it resolves to a static type

src/Microsoft.IIS.WebManager/app/main/main.ts
```diff
@@ -1,9 +1,29 @@
...
+import { enableProdMode } from '@angular/core'
+import { platformBrowserDynamic } from '@angular/platform-browser-dynamic'
 
-const platform = platformBrowserDynamic();
-platform.bootstrapModule(AppModule);
+if (environment.Production) {
+    enableProdMode();
+}
+
+if (environment.WAC) {
```
shhsu, line 15: Load different module depends on the environment

src/Microsoft.IIS.WebManager/app/main/polyfills.ts
```diff
@@ -0,0 +1,72 @@
+/**
```
shhsu, line 1: Standard polyfill stuff so the rest of the code would work

src/Microsoft.IIS.WebManager/app/resources/manifest.json
```diff
@@ -0,0 +1,181 @@
+{
```
shhsu, line 1: WAC needs this file

src/Microsoft.IIS.WebManager/app/resources/scripts/Get-Error.ps1
```diff
@@ -0,0 +1,5 @@
+Param([int]$wait)
```
shhsu, line 1: Brought in for WAC

src/Microsoft.IIS.WebManager/app/runtime/runtime.ts
```diff
@@ -0,0 +1,10 @@
+
```
shhsu, line 1: Runtime interface serves as entry point for any future WAC runtime specific logic including detect and install IIS administrator and automatically connect to hosts.

StandardRuntime is invoked for standard entrypoint and these operations would be no-op

src/Microsoft.IIS.WebManager/gulpfile.js
```diff
@@ -0,0 +1,128 @@
+"use strict";
```
shhsu, line 1: Our project would now build using this gulpfile which is standard for WAC extensions

src/Microsoft.IIS.WebManager/gulps/gulp-merge-json-in-folders/index.js
```diff
@@ -0,0 +1,57 @@
+/* tslint:disable */
```
shhsu, line 1: The gulps folder is brought in from WAC extensions

src/Microsoft.IIS.WebManager/tools/UpdateExtensionName.ps1
```diff
@@ -0,0 +1,75 @@
+﻿Param(
```
shhsu, line 1: The tools folder is brought in from WAC extensions

shhsu:
@jimmyca15 @bariscaglar @bangbingsyb 

High level of this change

Objectives

1. Have the project build and deploy with angular cli
- import function can no longer take expression. Angular cli runs a static analysis which require the parameter to be constant
- Implement .environment files for different compilation environment such as dev/prod/wac
- Added angular font-awesome package to support font awesome

2. Integrate web manager project as WAC extension
- Brought in WAC extension components such as gulp/tools folder and manifest.json
- Use the standard gulp file provided by WAC team to build web manager

3. WAC component was introduced but broke the build because our current code is implemented with older libraries and dependencies
- Ported ES6
- More rigorous type checking is needed
- Object such as timer is now persisted as an object

4. Introduce a Runtime interface that would serve as entry point for WAC specific logics such as WAC module initialization, and for the future it would contain logic to install IIS administrator on target server and automatically connect to the host WAC currently connected to

shhsu:
> 
> 
> Since this PR is so big do you plan on writing a description for all the changes, especially the architectural ones, contained in this?

commented above

jimmyca15:
@shhsu

I'd like to confirm some things

1. We're no longer able to build through VS?
2. We are still targeting ES5 for transpilation?
3. The app can still be hosted in IIS?

shirhatti:
> Added angular font-awesome package to support font awesome

We need to go through the OSS approval tool to pull in a new OSS dependency. @shhsu Can you start a thread with me?

shhsu:
@jimmyca15 
* Currently, no. We cannot build it in VS but we can support it if required
* Yes, still targeting ES5
* web.config need to be updated for this to host on iis. I do not plan to fix that on this PR yet.

@shirhatti 
Thanks for the reminder. I will start a new thread

src/Microsoft.IIS.WebManager/angular-cli.json
```diff
@@ -0,0 +1,49 @@
...
+      "template": false
+    },
+    "prefixInterfaces": false,
+    "styleExt": "css",
+    "build": {
+      "showCircularDependencies": true
+    }
+  },
+  "project": {
+    "name": "wac-iis",
```
jimmyca15: Isn't the project name IIS.WebManager?

src/Microsoft.IIS.WebManager/app/files/file-nav.service.ts
```diff
@@ -115,8 +116,7 @@ export class FileNavService implements IDisposable {
         //
         // Navigation
         this._defaultPath = defaultPath;
-        this._nav = new Navigator(this._route, this._location, useHash, defaultPath);
-
+        this._nav = new Navigator(this._route, this._location, useHash, defaultPath)
```
jimmyca15: Looks like a _;_ was removed here.

src/Microsoft.IIS.WebManager/app/main/app.component.ts
```diff
@@ -67,36 +68,40 @@ import { ServerAnalyticService } from '../webserver/server-analytic.service';
...
-                private _serverAnalyticService: ServerAnalyticService,
-                private _renderer: Renderer,
-                angulartics2: Angulartics2,
-                angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics) {
+        private _connectService: ConnectService,
+        private _loadingSvc: LoadingService,
+        private _windowService: WindowService,
+        private _versionService: VersionService,
+        private _serverAnalyticService: ServerAnalyticService,
+        @Inject("Runtime") private runtime: Runtime,
```
jimmyca15, line 76: I take it this comes from WAC. They prescribe us to destroy it when our component is removed?

src/Microsoft.IIS.WebManager/angular-cli.json
bariscaglar: I believe Jimmy is looking for a bit more detailed overview.

src/Microsoft.IIS.WebManager/app/main/app.component.ts
shhsu: This object is created by me. In the standard deployment scenario it is an empty object with a bunch of no-ops. In WAC scenario there are some operations that would take place including initialization their context and destroying

src/Microsoft.IIS.WebManager/angular-cli.json
shhsu: Sure, would make more sense.

src/Microsoft.IIS.WebManager/angular-cli.json
shhsu: Yes, I have put descriptions in subsequent posts

shhsu:
@jimmyca15 @bangbingsyb 
I just checked in a giant change and fixed any Ahead Of Time compilation error in angular cli prod mode.
It includes
1. more module path renaming in `webserver-routing.module.ts`
2. Apparently in AOT compilation, it doesn't allow private members of a class to be accessed in view. I have removed the private qualifier of a bunch of variable because of this

shhsu:
@shirhatti 
I have started a OSS review, ID 905306

src/Microsoft.IIS.WebManager/app/files/file-editor.ts
jimmyca15: Isn't this a behavioral change though? Previously the subject was able to emit null now it isn't. What if the user wants to delete all text then save?

src/Microsoft.IIS.WebManager/app/files/file-editor.ts
```diff
@@ -136,10 +141,12 @@ export class FileEditor {
             .catch(e => { this._unsupported = true });
     }
 
-    private codeEditorLoaded(lang: any) {
+    codeEditorLoaded(lang: any) {
         if (lang) {
             this.getFileText().then(txt => {
```
jimmyca15: If getFileText can return void, then calling `.then` could throw a nullref.

src/Microsoft.IIS.WebManager/app/files/file-explorer.ts
```diff
@@ -58,39 +58,39 @@ export class FileExplorer implements OnDestroy {
         this._subscriptions.forEach(sub => sub.unsubscribe());
     }
 
-    private refresh() {
+    refresh() {
```
jimmyca15: Why are we removing private?

src/Microsoft.IIS.WebManager/app/files/navigator.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable } from '@angular/core';
+import { Injectable, Optional } from '@angular/core';
```
jimmyca15: I see `Optional` was added, but I don't see it being used.

src/Microsoft.IIS.WebManager/app/main/app.component.ts
```diff
@@ -67,39 +68,43 @@ import { ServerAnalyticService } from '../webserver/server-analytic.service';
...
 
     ngOnDestroy() {
-        this._loadingSvc.destroy();
+        this._loadingSvc.destroy()
+        this.runtime.DestroyContext()
     }
 
     isRouteActive(route: string): boolean {
-        return this._router.isActive(route, true);
+        var isActive = this._router.isActive(route, true);
```
jimmyca15: Why don't we just return `this._router.isActive(...)`?

src/Microsoft.IIS.WebManager/app/main/home.component.ts
jimmyca15: There is a reason these were dynamic. Whenever a tab contains a dynamic component, the component is activated whenever the tab is selected. This allows for lazy loading of tabs. If this wasn't the case, then all the tabs in a tabs component would attempt to load their content and display at the same time. If there are a lot of tabs this can be quite heavy. I would recommend sticking with the dynamic.

[Dynamic component activate](https://github.com/Microsoft/IIS.WebManager/blob/cb66e033ccfbc3e75437e0450c33ed991dfd2a3b/src/Microsoft.IIS.WebManager/app/common/dynamic.component.ts#L40)
[Tab activate](https://github.com/Microsoft/IIS.WebManager/blob/cb66e033ccfbc3e75437e0450c33ed991dfd2a3b/src/Microsoft.IIS.WebManager/app/common/tabs.component.ts#L327)
[Tabs component calling activate](https://github.com/Microsoft/IIS.WebManager/blob/cb66e033ccfbc3e75437e0450c33ed991dfd2a3b/src/Microsoft.IIS.WebManager/app/common/tabs.component.ts#L262)

src/Microsoft.IIS.WebManager/app/main/settings.ts
```diff
@@ -0,0 +1,65 @@
...
+export const FilesComponentName = "FilesComponent"
+export const UploadComponentName = "UploadComponent"
+export const WarningComponentName = "WarningComponent"
+export const AppModuleName = "AppModule"
+
+export class ComponentReference {
+    constructor(public name: string, public ico: string, public component_name: string, public api_name: string, public api_path: string) {}
+}
+
+export const FilesComponentReference: ComponentReference = new ComponentReference("Files", "fa fa-files-o", FilesComponentName, "files", "/api/webserver/url-rewrite/{id}")
```
jimmyca15: Files component points to url rewrite?

src/Microsoft.IIS.WebManager/app/index.html
```diff
@@ -0,0 +1,30 @@
...
+    <meta http-equiv="pragma" content="no-cache" />
+    <meta http-equiv="expires" content="0" />
+    <meta http-equiv="X-UA-Compatible" content="IE=11" />
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; connect-src *; script-src 'self' 'unsafe-eval' www.google-analytics.com; img-src 'self' www.google-analytics.com iis.net; style-src 'self' 'unsafe-inline'; font-src 'self';">
+    <meta http-equiv="Strict-Transport-Security" content="max-age=86400">
+    <title>Microsoft IIS Web Manager</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link src="node_modules/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet" />
+    <link src="node_modules/font-awesome/css/font-awesome.min.css" rel="stylesheet" />
+    <link rel="stylesheet" src="themes/main.css" type="text/css" />
```
jimmyca15, line 17: why is css no longer being versioned?

src/Microsoft.IIS.WebManager/js/main.js
```diff
@@ -1,70 +0,0 @@
...
-                'angulartics2': 'node_modules/angulartics2',
-                'rxjs': 'node_modules/rxjs',
-                'ts': 'node_modules/plugin-typescript@4.0.10/lib/plugin.v' + ver('ts') + '.js',
-                'typescript': 'node_modules/typescript@2.0.2/lib/typescript.v' + ver('typescript') + '.js'
-            },
-            packages: {
-                defaultExtension: 'js',
-                app: {
-                    format: 'cjs',
-                    defaultExtension: 'v' + SETTINGS.version + '.js'
```
jimmyca15, line 54: We used system js configuration to version the application. This allowed us to break caches whenever we released an update. Where is this being done now?

src/Microsoft.IIS.WebManager/app/main/settings.ts
jimmyca15: What is this used for? There is a files componentreference in the GLOBAL_MODULES.

src/Microsoft.IIS.WebManager/app/notification/notification.component.ts
```diff
@@ -38,23 +36,24 @@ import { DynamicComponent } from '../common/dynamic.component';
         <div class="notifications shadow">
             <div *ngIf="_warning" class="entry warning border-active" (keyup.esc)="clearWarning()" tabindex="-1" autofocus (blur)="onBlur()" (focus)="onFocus()">
                 <i class="fa fa-times exit" (click)="clearWarning()" title="Dismiss"></i>
-                <dynamic [name]="_warning.componentName" [module]="_warning.module" [data]="_warning.data" [eager]="true"></dynamic>
+                <dynamic [name]="_warning.componentName" [module]="_warning" [data]="_warning.data" [eager]="true"></dynamic>
```
jimmyca15: `[module]` is a `ComponentReference`. `_warning` is a `Notification`. So why was .module was removed from _warning.

src/Microsoft.IIS.WebManager/app/notification/notification.component.ts
```diff
@@ -38,23 +36,24 @@ import { DynamicComponent } from '../common/dynamic.component';
...
             </div>
             <div [hidden]="!_active">
                 <div *ngFor="let notification of _notifications; let i = index;" [hidden]="notification._hidden" class="entry background-normal border-active">
                     <i class="fa fa-times exit" (click)="dismiss(i)" title="Dismiss"></i>
-                    <dynamic [name]="notification.componentName" [module]="notification.module" [data]="notification.data" [eager]="true"></dynamic>
+                    <dynamic [name]="notification.componentName" [module]="notification" [data]="notification.data" [eager]="true"></dynamic>
```
jimmyca15: Same as above.

src/Microsoft.IIS.WebManager/app/webserver/authentication/basic-auth.component.ts
```diff
@@ -41,12 +41,12 @@ import { NotificationService } from '../../notification/notification.service';
...
 })
 export class BasicAuthenticationComponent implements OnDestroy {
-    private _model: BasicAuthentication;
+    _model: BasicAuthentication;
     private _locked: boolean;
     private _original: BasicAuthentication;
     private _subscriptions: Array<Subscription> = [];
 
-    constructor(private _service: AuthenticationService,
+    constructor(public _service: AuthenticationService,
```
jimmyca15: Why public?

src/Microsoft.IIS.WebManager/app/webserver/url-rewrite/outbound-rules/outbound-rules.component.ts
```diff
@@ -10,7 +10,7 @@ import { OutboundSection, OutboundRule, PatternSyntax, OutboundTags, ActionType,
     selector: 'outbound-rules',
     template: `
         <error [error]="_service.outboundError"></error>
-        <div *ngIf="!_service.outboundError && _settings">
+        <div *ngIf="!_service.outboundError && _settings != null">
```
jimmyca15: If we're using _settings != null shouldn't we also use _service.outboundError == null?

src/Microsoft.IIS.WebManager/app/files/file-editor.ts
shhsu: That's right. Fixing it

src/Microsoft.IIS.WebManager/app/files/file-editor.ts
shhsu: should have fixed this issue in getFileText instead. This is done now

src/Microsoft.IIS.WebManager/app/files/file-explorer.ts
shhsu: angular AOT (ahead of time) compilation seems to disallow view components to access private methods/properties.

src/Microsoft.IIS.WebManager/app/files/navigator.ts
shhsu: neither was injectable, removed

src/Microsoft.IIS.WebManager/app/index.html
shhsu: Angular cli build process would insert a hash in a filenames if you build in prod mode

src/Microsoft.IIS.WebManager/app/main/app.component.ts
shhsu: This line was added during debugging process. no longer needed. removed

src/Microsoft.IIS.WebManager/app/main/home.component.ts
shhsu: Fair enough, this set changes was done in a very stage where I didn't know how to fix dynamic components for angular cli. I will change them back

src/Microsoft.IIS.WebManager/app/main/settings.ts
shhsu: Good catch, point to the wrong item.
This is used in webserver.component.ts. It's not a part of GLOBAL_MODULES

src/Microsoft.IIS.WebManager/app/notification/notification.component.ts
shhsu: right, that was an oversight

src/Microsoft.IIS.WebManager/app/notification/notification.component.ts
shhsu: right, that was an oversight

src/Microsoft.IIS.WebManager/app/webserver/authentication/basic-auth.component.ts
shhsu: AOT compilation issue as it is used in view

src/Microsoft.IIS.WebManager/app/webserver/url-rewrite/outbound-rules/outbound-rules.component.ts
shhsu: This was added during debugging. Should not be needed. Removing

jimmyca15:
@shhsu I just saw the reason you removed private. That is a huge change. Is there any way to fix this? This is not something we should just remove, it is an integral part of the code.

shhsu:
> @shhsu I just saw the reason you removed private. That is a huge change. Is there any way to fix this? This is not just something we should remove, it is an integral part of the code.

Do you have concerns about functional bugs or it's cleanliness issue. Potentially we can create many public getters to fix this issue.

shhsu:
Also, if we are not using AOT at all and never optimize our code. it would just run fine with private

jimmyca15:
@shhsu I am not concerned about current functional bugs because of removing privates. I am concerned by the design that is promoted and possible future bugs. The private vs public access modifiers allow us to create properly encapsulated types. If we just remove private everything will definitely work, but it totally destroys the design.

Is it possible not to use AOT?

shhsu:
@jimmyca15 

In this case I will create public getters. I will update the PR once I am done

src/Microsoft.IIS.WebManager/js/main.js
shhsu: Angular cli injects a hash on file names when compile to prod

shhsu:
@jimmyca15 

I have checked in a change that make variables private

src/Microsoft.IIS.WebManager/app/certificates/certificates-list.component.ts
```diff
@@ -121,6 +120,10 @@ export class CertificatesListComponent implements OnInit, OnDestroy {
         }
     }
 
+    get serviceLoading() {
```
jimmyca15, line 123: If we're adding back privates because we aren't using AOT, then we don't need these getters which are just fluff at this point.

src/Microsoft.IIS.WebManager/app/certificates/certificates-list.component.ts
```diff
@@ -159,7 +162,7 @@ export class CertificatesListComponent implements OnInit, OnDestroy {
         this.onRangeChange(this._range);
     }
 
-    private refresh() {
+    refresh() {
```
jimmyca15: Private was removed?

src/Microsoft.IIS.WebManager/app/header/header.component.ts
```diff
@@ -143,13 +143,17 @@ export class HeaderComponent implements OnDestroy {
...
                     clearTimeout(this._timeout);
-                    this._timeout = 0;
+                    this._timeout = null;
                     setTimeout(_ => this._inProgress = false, 300);
                 }
             }
         }));
     }
 
+    private get options() {
```
jimmyca15, line 153: Why do we need this? It's private, the field is private. Why not just access the field?

src/Microsoft.IIS.WebManager/app/main/app.component.ts
```diff
@@ -1,4 +1,4 @@
-import { NgModule, Component, ViewEncapsulation, OnInit, OnDestroy, ViewChild, ElementRef, Renderer } from '@angular/core';
+import { Inject, Component, ViewEncapsulation, OnInit, OnDestroy, Optional, ViewChild, ElementRef, Renderer } from '@angular/core';
```
jimmyca15: I don't see `Optional` being used anywhere.

src/Microsoft.IIS.WebManager/app/main/home.component.ts
```diff
@@ -21,25 +25,31 @@ import { OptionsService } from './options.service';
...
 export class HomeComponent {
-    constructor(private _options: OptionsService,
-        private _route: ActivatedRoute) {
+    WebSiteListComponentReference = new ComponentReference("WebSiteListComponent", null, WebSiteListComponentName, null, null)
+    FilesComponentReference = new ComponentReference("FilesComponent", null, FilesComponentName, null, null)
+    MonitoringComponentReference = new ComponentReference("MonitoringComponent", null, MonitoringComponentName, null, null)
+    constructor(private _options: OptionsService) {
+    }
+
+    get options() {
```
jimmyca15, line 52: Unnecessary getter addition?

src/Microsoft.IIS.WebManager/app/settings/settings.component.ts
```diff
@@ -25,7 +25,10 @@ import { OptionsService } from '../main/options.service';
     `]
 })
 export class SettingsComponent {
-
     constructor(private _options: OptionsService) {
     }
+
+    get options() {
```
jimmyca15, line 31: Unnecessary getter addition?

src/Microsoft.IIS.WebManager/app/webserver/authorization/authorization.component.ts
```diff
@@ -58,6 +58,10 @@ export class AuthorizationComponent implements OnInit, OnDestroy {
         this._subscriptions.forEach(sub => sub.unsubscribe());
     }
 
+    get service() {
```
jimmyca15, line 61: Unnecessary getter addition?

src/Microsoft.IIS.WebManager/app/certificates/certificates-list.component.ts
shhsu: I didn't want to remove this. Eventually we would move to AOT. It's not hurting anything.

src/Microsoft.IIS.WebManager/app/header/header.component.ts
shhsu: same as above

jimmyca15 (approved):

PR #216, opened 2018-10-11
Add HSTS

shirhatti:


shhsu (approved):

jimmyca15:
Why don't we just add the header to the custom header list instead of adding a url rewrite rule? The URL rewrite rule is kind of complicated for just adding a header. I see that we are only adding it for HTTPS requests, but there should be no issue sending it over HTTP. In fact the HSTS doc mentions this.

```<?xml version="1.0"?>
<configuration>
  <system.webServer>
    <httpProtocol>
      <customHeaders>
        <add name="X-Frame-Options" value="SAMEORIGIN" />
        <add name="Strict-Transport-Security" value="max-age=31536000; includeSubDomains" /> 
      </customHeaders>
    </httpProtocol>
```

shirhatti:
Because the HSTS spec forbids you from sending the HSTS header over HTTP

shirhatti:
See section 7.2- https://tools.ietf.org/html/rfc6797#section-7.2

jimmyca15:
Thank you for the link. It is actually an absurd requirement that doesn't make much sense. HTTP response should be out of scope for HSTS. HSTS is to protect man in the middle. In HSTS scenarios assume the server is already under attack. Man in the middle can hijack HTTP response and add/remove HSTS header at will. 

In any case, I am not one to fight against RFC.

jimmyca15 (approved):

PR #226, opened 2018-11-26
WAC integration functionally complete

shhsu:


src/angular-cli.json
```diff
@@ -42,7 +42,7 @@
     "prefixInterfaces": false,
     "styleExt": "css",
     "build": {
-      "showCircularDependencies": true
+      "showCircularDependencies": false
```
shhsu, line 45: dynamicComponents nesting (webserver component for example) introduced circular dependencies. We can't avoid it. Muting warning messages

src/app/common/http-facade.ts
```diff
@@ -0,0 +1,7 @@
+import { Request, Response } from "@angular/http"
+import { Observable } from "rxjs/Observable";
+
+export interface HttpFacade {
```
shhsu, line 4: Normal website path: uses angular/http classes to make requests again IIS Admin host
WAC path: uses WAC/Powershell to invoke local_http.ps1 and make local http request on IIS admin host isself

shhsu:
Overview of this PR:

* Added installer page for WAC scenario. User will be prompt to install IIS Admin API of target host does not have one
* Some UI fixes, including adding IdleComponent as WAC will navigate to idle before navigating away
* Code clean up: remove unused using, styles, avoid calling toPromise multiple time in a function or code path, move static imports and function imports to the end of imports
* Properly clean up api tokens created by WAC
* Instead of adding and modifying routing for WAC scenario, create 2 separate routing classes
* Fixed a bug that in angular 4+, ActiveRoute object can have empty fields if it was unchanged since last invocation.

shhsu:
@mattatmsft

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
+        let req = new Request(reqOpt);
         //
         // Set Access-Token
-        req.headers.set('Access-Token', 'Bearer ' + this._conn.accessToken);
-        
-        return this._http.request(req).toPromise()
-            .catch(e => {
+        req.headers.set('Access-Token', 'Bearer ' + conn.accessToken);
+        return this._http.request(req)
+            .catch((e, c) => {
```
jhkimnew: variable c is not actually use. Please use it or remove. 

FYI, 
When Observable is used for retrying, it had to return a new observable object in the .catch block. If our intention is not the case (no retry), I think we should just throw error like the below example. Please check if this can be used here. :-)

```
    of(1, 2, 3, 4, 5)
      .map(n => {
      if (n == 4) {
        throw 'four!';
      }
      return n;
    })
    .catch(err => Observable.throw(err))
    .subscribe(x => console.log(x));
```

shhsu:
Added NuGet packaging

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
-                            return this._http.options(this._conn.url).toPromise()
-                                .catch(e => {
-                                    this._connectSvc.reconnect();
-                                    return Promise.reject("Not connected");
+                if (e instanceof Response) {
+                    if (e.status == 0) {
+                        return this._http.request(req)
+                            .catch((err, _) => {
+                                // Check to see if connected
+                                return this._http.options(this._conn.url)
```
jhkimnew, line 134: previously, we handled error case and success case. but you implemented only error case. Please add the success case as we did previously.

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
+                                    .catch((e, c2) => {
+                                        this._connectSvc.reconnect();
+                                        this.handleHttpError(err);
+                                        return c2
```
jhkimnew: When error happens, returning c2 seems not make sense. Please check if we can only throw error.

build.ps1
```diff
@@ -24,6 +24,8 @@ try {
 
     Write-Host "Building project..."
     gulp build $buildArgs
+
+    nuget pack -OutputDirectory ..\dist
```
jhkimnew: if nuget.exe is not in Path, this won't work. Please add error handling or update build instruction so that users update Path environment variable to include the nuget.exe path.

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
-
-    public request(url: string, options?: RequestOptionsArgs, warn?: boolean): Promise<any> {
-        if (!this._conn) {
-            this._connectSvc.gotoConnect(true);
-            return Promise.reject("Not connected");
-        }
-
-        let req: Request;
-
+    private performRequest(conn: ApiConnection, url: string, options?: RequestOptionsArgs, warn?: boolean): Observable<Response> {
```
jhkimnew, line 111: please move performRequest() below requestOnConnection() considering the function relation is like "request() -> requestOnConnection() -> performRequest()".

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
+    public request(url: string, options?: RequestOptionsArgs, warn?: boolean): Promise<Response> {
+        return this.requestOnConnection(url, options, warn).toPromise()
     }
 
-    private handleHttpError(err, warn?: boolean): Promise<any> {
+    private requestOnConnection(url: string, options?: RequestOptionsArgs, warn?: boolean): Observable<Response> {
+        if (this._conn) {
+            return this.performRequest(this._conn, url, options, warn)
+        } else {
+            return this.runtime.ConnectToIISHost().mergeMap(c => {
```
jhkimnew: please rename c with conn for readability

src/app/common/httpclient.ts
```diff
@@ -108,68 +95,72 @@ export class HttpClient {
...
+                return this.performRequest(c, url, options, warn)
+            })
+        }
+    }
+
+    private handleHttpError(err, warn?: boolean) {
         if (err.status == 403 && err.headers.get("WWW-Authenticate") === "Bearer") {
             this._connectSvc.reconnect();
-            return Promise.reject("Not connected");
+            throw "not connected"
```
jhkimnew: not => Not

src/app/connect/connect.component.ts
```diff
@@ -201,13 +201,9 @@ export class ConnectComponent implements OnDestroy {
...
     private save() {
         this._service.save(this._conn);
-
-        this._service.active.subscribe(c => {
-            if (!c) {
-                this._service.connect(this._conn);
-            }
-        }).unsubscribe();
-
+        var sub = this._service.connect(this._conn).subscribe(_ => {}, _ => {}, () =>{
```
jhkimnew: previously, it seems to reconnect when connection is not established. But you did not implement that. Is this by design?

src/app/connect/connect.service.ts
```diff
@@ -71,15 +70,18 @@ export class ConnectService {
...
                 this.complete(conn);
-                return Promise.resolve(conn);
+                this.save(conn);
+                return conn;
             })
-            .catch(_ => {
-                this.gotoConnect(true);
-
-                return this.ping(conn, new Date().getTime() + ConnectService.PING_TIMEOUT, popup);
+            .catch((e, __) => {
```
jhkimnew: I am wondering if we can refactor (e, __)  ==> (err, caught) and reuse caught instead of creating a new Observable object. Can you check that?

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -0,0 +1,59 @@
+param(
+    [string]
+    $requestBase64
+)
+
+$ErrorActionPreference = "Stop"
+$contentEncoding = [System.Text.Encoding]::UTF8
+
+function stringify($content) {
+    if ((!$content) -or ($content -is [System.String])) {
```
jhkimnew, line 10: Can we refactor this as the following in order to remove else?

```
    if ((!$content) -or ($content -is [System.String])) {
        return $content
    } 

    if ($content -is [System.Byte[]]) {
        return [System.Convert]::ToBase64String($content)
    } 
    return $content.ToString()
    }
```

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+try {
+    $res = Invoke-WebRequest -UseBasicParsing -UseDefaultCredentials @req
+} catch {
+    $res = $_.Exception.Response
+} finally {
+    if (!$res) {
+        throw "No response is returned."
```
jhkimnew: When exception happens, I think we should pass the error code and the url information instead of returning "No response is returned." so that users can troubleshoot.  As far as I remember, invoke-webrequest throw exception for permisson error or 5xx error.  The detailed error can be helpful to figure out what error happened as the below example of bogus request:
```
PS C:\> try { invoke-webrequest http://localhost:1234/bogus } catch { $_.Exception; $_.Exception.InnerException; }
Unable to connect to the remote server
No connection could be made because the target machine actively refused it 127.0.0.1:1234
```

src/app/resources/iis_scripts/token_utils.ps1
```diff
@@ -0,0 +1,89 @@
...
+        "ContentType" = "application/json";
+        "WebSession" = $sess
+    }
+    if ($subpath) {
+        $requestParams.Uri += "/${subpath}"
+    }
+    if ($requestBody) {
+        $requestParams.Body = (ConvertTo-Json -Compress $requestBody)
+    }
+    $tokenUpsert = Invoke-WebRequest @requestParams
```
jhkimnew, line 57: What Upsert means? Typo?

src/app/resources/iis_scripts/token_utils.ps1
```diff
@@ -0,0 +1,89 @@
...
+$RenewEndpoint = "access-tokens"
+$CreateEndpoint = "api-keys"
+$DeleteEndpoint = "api-keys"
+function VerifyResponse([string] $action, [Microsoft.Powershell.Commands.WebResponseObject] $response) {
+    if ($response.StatusCode -ge 300) {
+        throw "Invalid status code $($response.StatusCode) while performing action: $action"
+    }
+}
+function TokenRequest([string] $targetEndpoint, [string]$method, [string]$subpath, $requestBody) {
+    $sessionCreate = Invoke-WebRequest "$apiHost/security/$targetEndpoint" -UseBasicParsing -UseDefaultCredentials -SessionVariable sess
```
jhkimnew, line 36: I think we should catch exception for all the Invoke-WebRequest to get the error information to help users to troubleshoot the problem.

src/app/resources/iis_scripts/token_utils.ps1
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+    VerifyResponse "query exsiting tokens" $query | Out-Null
+    $existingToken = (ConvertFrom-Json $contentEncoding.GetString($query.Content)).api_keys | Where-Object { $_.purpose -eq $tokenName }
+}
+
+if ($command -eq 'ensure') {
+    if ($existingToken) {
+        ## always renew token when this script is called because there was no way to query for existing token's value
+        $existingToken.expires_on = (Get-Date).AddDays(3).ToString()
```
jhkimnew: Please define a variable for the default expiration day for the 3 days to avoid hard-coded value.

src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -0,0 +1,83 @@
...
+      PowerShellScripts.local_http,
+      { requestBase64: requestEncoded },
+      (k, v) => {
+        if (k === "body") {
+          try {
+            return atob(v)
+          } catch {
+            return v
+          }
+        } else if (k === "headers") {
```
jhkimnew: remove "else" for readability

src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -0,0 +1,83 @@
...
+          // we need to explicitly wrap it otherwise when we pass it to new Response(res), the header would remain a plain object
+          return new Headers(v)
+        }
+        return v
+      }).map(res => {
+        let response = new Response(res)
+        if (res.status < 200 || res.status >= 400) {
```
jhkimnew, line 59: In this condition, 3xx status code are considered as good as 2xx. Is this by design? I am wondering if IIS.Administration API returns any 3xx status code. Please check that. If not, I think 3xx status code should be considered as error case.

src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -0,0 +1,83 @@
...
+    })
+  }
+
+  private invoke<T>(pwCmdString: string, psParameters: any, reviver: (key: any, value: any) => any = null): Observable<T> {
+    var compiled = PowerShell.createScript(pwCmdString, psParameters)
+    var name = pwCmdString.split('\n')[0]
+    return this.session.mergeMap(ps => {
+      return ps.powerShell.run(compiled).mergeMap(response => {
+        if (!response) {
+          throw `Powershell command ${name} returns no response`;
```
jhkimnew, line 72: we should gather the detailed error message and include that here as I mention in the comment of the powershell script.

build.ps1
jhkimnew: The output directory is misleading users because it uses relative path. Please use the full path to avoid any confusion.

jhkimnew: I think "nuget pack" should be called only when the --wac is set because the nuget package is only used by WAC.

shhsu: added more input parameters and checking logic that should resolve these issues (next commit)

src/app/common/httpclient.ts
shhsu: moved

src/app/common/httpclient.ts
shhsu: Fixed as discussed

src/app/common/httpclient.ts
shhsu: renamed (next commit)

src/app/common/httpclient.ts
shhsu: updated

src/app/connect/connect.component.ts
shhsu: The previous code is just a roundabout way of doing what I wrote. since the subscriber would always see this._conn

However, I am reverting this line since it is out of scope of this PR

src/app/connect/connect.service.ts
shhsu: we dont' have to, single underscore would work. Changed

src/app/resources/iis_scripts/local_http.ps1
shhsu: Ok

src/app/resources/iis_scripts/local_http.ps1
shhsu: I did assign the response object in the catch block though. The !$res scenario should never happen since non-success status code would be captured. In fact I am using the error status code on my connection logic.
This line is only here so there wouldn't be any NPE but if you do reach this line it's a completely unexpected error. Not HTTP error

src/app/resources/iis_scripts/token_utils.ps1
shhsu: Wouldn't the error be thrown upstream?

src/app/resources/iis_scripts/token_utils.ps1
shhsu: Upsert => Insert and/or Update

src/app/resources/iis_scripts/token_utils.ps1
shhsu: done

src/app/runtime/wac/services/powershell-service.ts
shhsu: changed to switch statement

src/app/runtime/wac/services/powershell-service.ts
shhsu: it doesn't. Usually 3XX is successfully though  just some more work need to be done on client side

src/app/runtime/wac/services/powershell-service.ts
shhsu: This response object is not the http response. It's WAC's powershell service's response. The HttpResponse returned is wrapped within response.results which I did process underneath.

src/app/resources/iis_scripts/local_http.ps1
jhkimnew: The catch statement won't work. Please run below commands and you can see the $res is still $null value. 

```
$res = $null; 
try { 
    invoke-webrequest http://localhost:1234/bogus;
} 
catch { 
    $res = $_.Exception.Response 
}; 
$res
```

shhsu: The status code 4XX and 5XX should already work. In fact I think all the non zero status code scenario works.
However, I think this bring up a good point. I have modified the script to return the error message when status code is 0 in this revision

src/app/common/httpclient.ts
jhkimnew: Success case is not handled yet.

src/app/common/httpclient.ts
```diff
@@ -108,68 +99,67 @@ export class HttpClient {
...
+                return Observable.throw(apiError)
+            })
     }
 
-    private handleHttpError(err, warn?: boolean): Promise<any> {
+    private handleHttpError(err, warn?: boolean) {
         if (err.status == 403 && err.headers.get("WWW-Authenticate") === "Bearer") {
             this._connectSvc.reconnect();
-            return Promise.reject("Not connected");
+            throw "Not connected"
```
jhkimnew, line 158: I guess this should Observable.throw. Would you test this scenarios and confirm which one is better between throw and Observable.throw?

jhkimnew (approved):
@shhsu: one update here. The default permission to the inetpub is read-only. For example, if you try to create a new file under the inetpub, you will get permission error.

Looks good. We need to test this on the downlevel OS as well. And please make sure opening a new issue that users uses only allowed to inetpub directory. We also need to consider if there is any security issue on the default setting which allows read permission to the c:\inetpub directory.

shhsu:
thanks, creating work items

PR #230, opened 2018-12-06
Fixed accessibility issues

jhkimnew:
Added autofocus and fix the tab order with switching from (actions, UI controls) to (UI controls, actions).

NOTE:

1. In order to make the autofocus directive works, the element should be re-generated according to the value *ngIf directive. So, I had to add some *ngIf directives when required.

2. We still need to show the actions buttons on top of the form even though the actions buttons are moved bottom. In order to do that, I made one table (1 * 2) and make the first column uses 8/12 and uses the remained (4/12) are used for the action buttons.

3. Added keyup.[space|esc|enter] event handler to toggle/close/select focused element. 

4. .left-with-button is added to layout the text input box and then the button on the right side of the text input box.

5. In modules feature page, I added a wrapper span for switch button to make it autofocused. This is a work-around way and I will revisit to check if there is better solution.

### DIFF
--- a/src/app/common/focus.ts
+++ b/src/app/common/focus.ts
@@ -12,8 +12,10 @@ export class AutoFocus {
     }    
 
     ngOnInit() {
-        this._renderer.invokeElementMethod(this._el.nativeElement, 'focus', []);
-
+        setTimeout(() => {
+            this._renderer.invokeElementMethod(this._el.nativeElement, 'focus', []);
+        }, 1);
+        
         if (this._el.nativeElement.select) {
             setTimeout(() => {
                 this._renderer.invokeElementMethod(this._el.nativeElement, 'select', []);
--- a/src/app/files/file-list.ts
+++ b/src/app/files/file-list.ts
@@ -68,7 +68,8 @@ import { FileNavService } from './file-nav.service';
                     (dragstart)="drag(child, $event)"
                     (drop)="drop($event, child)"
                     (dragenter)="onDragItemEnter(child, $event)"
-                    (dragleave)="onDragItemLeave(child, $event)">
+                    (dragleave)="onDragItemLeave(child, $event)"
+                    (keyup.space)="select(child)">
                     <file [model]="child" (modelChanged)="doSort()"></file>
                 </li>
             </virtual-list>
@@ -404,6 +405,11 @@ export class FileListComponent implements OnInit, OnDestroy {
         else if (e.dataTransfer.effectAllowed == "all") {
             e.dataTransfer.dropEffect = "copy";
         }
+    }
+
+    private select(f) {
+        this.clearSelection();
+        this._selected.push(f);
     }
 
     private drag(f: ApiFile, e: DragEvent) {
--- a/src/app/webserver/central-certificates/central-certificate.component.ts
+++ b/src/app/webserver/central-certificates/central-certificate.component.ts
@@ -23,10 +23,8 @@ import { CentralCertificateConfiguration } from './central-certificates';
         <div *ngIf="_configuration">
             <fieldset class="path">
                 <label>Physical Path</label>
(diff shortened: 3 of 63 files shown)


src/app/common/focus.ts
```diff
@@ -12,8 +12,10 @@ export class AutoFocus {
     }    
 
     ngOnInit() {
-        this._renderer.invokeElementMethod(this._el.nativeElement, 'focus', []);
-
+        setTimeout(() => {
+            this._renderer.invokeElementMethod(this._el.nativeElement, 'focus', []);
+        }, 1);
```
shhsu, line 17: what is the purpose of the delay?

src/app/files/file-list.ts
```diff
@@ -68,7 +68,8 @@ import { FileNavService } from './file-nav.service';
                     (dragstart)="drag(child, $event)"
                     (drop)="drop($event, child)"
                     (dragenter)="onDragItemEnter(child, $event)"
-                    (dragleave)="onDragItemLeave(child, $event)">
+                    (dragleave)="onDragItemLeave(child, $event)"
+                    (keyup.space)="select(child)">
```
shhsu, line 72: Processing space key to select - is this a well understood feature or should we show as tooltip?

src/app/webserver/central-certificates/central-certificate.component.ts
```diff
@@ -23,10 +23,8 @@ import { CentralCertificateConfiguration } from './central-certificates';
         <div *ngIf="_configuration">
             <fieldset class="path">
                 <label>Physical Path</label>
-                <button title="Select Folder" [class.background-active]="fileSelector.isOpen()" class="right select" (click)="fileSelector.toggle()" [attr.disabled]="isPending || null" ></button>
-                <div class="fill">
```
shhsu, line 27: What's the purpose of removing this class

src/app/webserver/modules/module.component.ts
```diff
@@ -29,25 +29,19 @@ import {LocalModule, GlobalModule} from './modules';
         .visible-xs-inline-block {
             margin-left: 15px;
         }
+
+        .switch-wrapper {
+            // outline-style: dashed;
```
shhsu: removed unused

shhsu (changes requested):

src/app/common/focus.ts
jhkimnew: This is simply asynchronous call, which is common pattern(recommended) to call focus method from this kind of directive. I confirmed this works and there is no additional delay time because of this change.

src/app/files/file-list.ts
jhkimnew: Good point. Usually space key used to select focused item. So, users will figure out that they can use the space key for this purpose. 

Ideally, we should support arrow key instead of tab key for list view control. Once we support the arrow key, we can replace space key with right arrow key in the future, supposing the focused item is shown like "> mydir". The extra ">" indicates that the right arrow key can be used to select the current item.  I have opened this issue about that. 

https://github.com/Microsoft/IIS.WebManager/issues/231

src/app/webserver/central-certificates/central-certificate.component.ts
jhkimnew: Previously, the fill class was used to fill the space between the input text box and the file selector button becuase the file selector button was placed before the edit text box.

Now, that was fixed with switching the order and the button is supposed to be put right after the edit box automatically. 

The fill class can't be used anymore because the edit box will occupy whole line which will make the button is displayed in the next line. That's why I had to remove the fill class.

src/app/webserver/modules/module.component.ts
jhkimnew: Fixed.

shhsu (approved):

PR #233, opened 2018-12-09
Fix administrators permission issue

shhsu:
This PR makes it so we properly set up user permissions after installing IIS admin api.

The script now creates a group "WAC-IISAdmin" and ensure any WAC administrator is in the group.

"WAC-IISAdmin" group would also take ownership of IIS admin API installation directory


src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,120 @@
...
+    [ValidateSet('install', 'permission-check')]
+    [string]
+    $command,
+
+    [string]
+    $downloadFrom,
+
+    ## Not doing anything with this parameter yet
+    [string]
+    $sessionId,
```
jimmyca15, line 12: If it's not used, why have it?

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,120 @@
...
+
+    [string]
+    $serviceName = "Microsoft IIS Administration",
+
+    [string]
+    $iisAdminGroup = "WAC-IISAdmin"
```
jimmyca15: The name doesn't necessarily need WAC in it. Also, I would stay away from abbreviations in group names. The members of this group are being added to the "Owners" role within the IIS Administration API, so they will have the highest rights with the API. I would suggest naming it _IIS Administration Owners_.

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,120 @@
...
+        Start-Sleep -Seconds $waitPeriod
+        $remainingTime -= $waitPeriod
+        if ($remainingTime -lt 0) {
+            throw "timeout waiting for service to start"
+        }
+    }
+}
+
+function EnsureAdminGroup($user) {
+    New-LocalGroup -Name $iisAdminGroup -ErrorAction SilentlyContinue | Out-Null
```
jimmyca15: What if the group already exists?

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,120 @@
+param(
+    [Parameter(Mandatory=$true)]
+    [ValidateSet('install', 'permission-check')]
```
jimmyca15: _permission-check_ is unused.

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,120 @@
...
+
+$configLocation = [System.IO.Path]::Combine($workingDirectory, "config", "appsettings.json")
+$config = Get-Content -Raw -Path $configLocation | ConvertFrom-Json
+
+if ($install) {
+    $config.security.users.administrators = @( $iisAdminGroup )
+    $config.security.users.owners = @( $iisAdminGroup )
+    $updateConfig = $true
+} else {
+    if (!$config.security.users.administrators.Contains($user)) {
```
jimmyca15: I don't understand why this logic would be different whether installing or not. It seems like it should just be checks to see if the desired principal exists in _owners_ or _administrators_ and then add it.

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: For diagnostics purposes. I haven't decided how to use it yet.

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: sure

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: It's used in runtime.wac.ts

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: That's why I have SilentlyContinue set to make the code shorter.

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: I am assuming if the Admin API was already installed. I would like the original owners/administrators still have access to have API without using our groups

jimmyca15: I don't think that ensuring that the group that we create is part of the owners and administrators roles would remove any existing access of the installation. The WAC scenario would be more consistent if the IIS Administrator Owners group was added to the iis administration API roles regardless of whether the API was just installed or not.

src/app/resources/iis_scripts/admin_api_util.ps1
jimmyca15: What I meant was that it's not used in the script itself.

shhsu: Yeah, permission-check is just there as a enum value currently mean "not install". We might find that we need to do a different thing in the future.

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,126 @@
...
+
+$user = $(whoami)
+EnsureGroup $iisAdminOwners
+EnsureMember $iisAdminOwners $user
+
+if (!$config.security.users.owners.Contains($iisAdminOwners)) {
+    foreach ($member in $config.security.users.owners) {
+        EnsureMember $iisAdminOwners $member
+    }
+    $config.security.users.owners = @( $iisAdminOwners )
```
jimmyca15: Owner's could already have members right? Shouldn't $iisAdminOwners be appended to the existing list of owners rather than replace it.

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: I actually change the code to check for group existence and membership before creating or modifying. Because on second thought it would be more readable

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: Sure, I was kind of wondering between the two approaches. This may be cleaner. We should just communicate to clients that if WAC is involved we always manage permission with local groups

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,126 @@
...
+    $dirAccessGranted = $dirAcl.Access | Where-Object {(($_.IdentityReference.Value -eq "${env:ComputerName}\${iisAdminOwners}") -and (($_.FileSystemRights -bAnd $rights) -eq $rights))}
+    if (!$dirAccessGranted) {
+        $inheritFlags = [System.Security.AccessControl.InheritanceFlags]::ContainerInherit `
+            -bOr [System.Security.AccessControl.InheritanceFlags]::ObjectInherit
+        $propFlags = [System.Security.AccessControl.PropagationFlags]::InheritOnly
+        $ar = New-Object System.Security.AccessControl.FileSystemAccessRule($iisAdminOwners, $rights, $inheritFlags, $propFlags, "allow")
+        $dirAcl.SetAccessRule($ar) | Out-Null
+        $ar = New-Object System.Security.AccessControl.FileSystemAccessRule("BUILTIN\Administrators", $rights, $inheritFlags, $propFlags, "allow")
+        $dirAcl.SetAccessRule($ar) | Out-Null
+        $dirAcl.SetOwner([System.Security.Principal.NTAccount]$iisAdminOwners) | Out-Null
```
jimmyca15: Setting the owner of the IIS Administration API files to IIS Administration Owners? I wouldn't think we would want to go any less than administrators.

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: My current thinking is that if WAC get involved after Admin API is installed, we would just manage permission with local group. Note that there's a loop adding all the members in the iisAdminOwners group

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: right, this line was adding when I was struggling to find the proper acl. Turns out that was not needed. I have removed the line and leave SYSTEM as the user and it still works.

src/app/resources/iis_scripts/admin_api_util.ps1
jimmyca15: The issue i'm talking about is we are setting the owners role to __only__ IIS Administration Owners, when there could already be owners configured by the user. We should keep those. They aren't ours to delete.

shhsu: I am trying to keep a unified place to manage access. Consider this scenario: User A installed the program prior to using WAC. Now A is a owner in the config file but he has no write access to the config file as intended.
Now A uses WAC and the script is run, we still have to add user A to the iisAdminOwners for him to add iisAdminOwners into the owner group of the config file. Now this user's access would have been defined in 2 places, the config file says A has access. Also config file says iisAdminOwner has access, which he is.
Seems to be a problem in maintenance.

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -81,45 +81,39 @@ $configLocation = [System.IO.Path]::Combine($workingDirectory, "config", "appset
...
-        $inheritFlags = [System.Security.AccessControl.InheritanceFlags]::ContainerInherit `
-            -bOr [System.Security.AccessControl.InheritanceFlags]::ObjectInherit
-        $propFlags = [System.Security.AccessControl.PropagationFlags]::InheritOnly
-        $ar = New-Object System.Security.AccessControl.FileSystemAccessRule($iisAdminOwners, $rights, $inheritFlags, $propFlags, "allow")
-        $dirAcl.SetAccessRule($ar) | Out-Null
-        $ar = New-Object System.Security.AccessControl.FileSystemAccessRule("BUILTIN\Administrators", $rights, $inheritFlags, $propFlags, "allow")
-        $dirAcl.SetAccessRule($ar) | Out-Null
+    if (!$config.security.users.owners.Contains($iisAdminOwners)) {
+        if (!(Get-LocalGroupMember -Group "Administrators" -Member $user)) {
+            throw "Administrator privilege is needed ot initiate IIS Administration API"
```
jimmyca15: Administrator privilege is needed ot initiate IIS Administration API

->

Administrator privilege is needed **to** initiate IIS Administration API

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -0,0 +1,119 @@
...
+
+    ## Not doing anything with this parameter yet
+    [string]
+    $sessionId,
+
+    [string]
+    $serviceName = "Microsoft IIS Administration",
+
+    [string]
+    $iisAdminOwners = "IISAdminAPIOwners"
```
jimmyca15: We shouldn't use abbreviation in our group names.

This should be "IIS Administration API Owners"

jimmyca15 (approved):

PR #248, opened 2018-12-21
Use netstandard http client (System.Net.Http)

shhsu:
Invoke-WebRequest intermittently respond with "Server committed a protocol violation"

### DIFF
--- a/src/app/resources/iis_scripts/local_http.ps1
+++ b/src/app/resources/iis_scripts/local_http.ps1
@@ -34,9 +34,8 @@ $uriBuilder = [System.UriBuilder]$reqObj.url
 $uriBuilder.Host = "localhost"
 $uri = $uriBuilder.ToString()
 
-Write-Host $decoded
 try {
-    $httpMsg = New-Object System.Net.Http.HttpRequestMessage -ArgumentList $httpMethod,$uri
+    $httpMsg = New-Object System.Net.Http.HttpRequestMessage -ArgumentList $httpMethod, $uri
     if ($reqObj.Body) {
         $httpMsg.Content = $requestContent
         $requestContent = New-Object System.Net.Http.StringContent ([string] $reqObj.Body)
--- a/src/app/resources/iis_scripts/local_http.ps1
+++ b/src/app/resources/iis_scripts/local_http.ps1
@@ -4,70 +4,83 @@ param(
 )
 
 $ErrorActionPreference = "Stop"
+Add-Type -AssemblyName System.Net.Http
 $contentEncoding = [System.Text.Encoding]::UTF8
 
-function stringify($content) {
-    if ((!$content) -or ($content -is [System.String])) {
+function stringify([System.Byte[]]$content) {
+    if (!$content) {
         return $content
     }
-
-    if ($content -is [System.Byte[]]) {
-        return [System.Convert]::ToBase64String($content)
-    }
-
-    return $content.ToString()
+    return [System.Convert]::ToBase64String($content).ToString()
 }
 
 ## same order as @angular/http/enums/RequestMethod
+## Note that index 0 maps to GET, so if $reqObj.method is null we would default to GET
 $requestMethods = @(
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Get,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Post,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Put,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Delete,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Options,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Head,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Patch
+    [System.Net.Http.HttpMethod]::Get,
+    [System.Net.Http.HttpMethod]::Post,
+    [System.Net.Http.HttpMethod]::Put,
+    [System.Net.Http.HttpMethod]::Delete,
+    [System.Net.Http.HttpMethod]::Options,
+    [System.Net.Http.HttpMethod]::Head,
+    [System.Net.Http.HttpMethod]::Patch
 )
 
 $decoded = $contentEncoding.GetString([System.Convert]::FromBase64String($requestBase64))
 $reqObj = ConvertFrom-Json $decoded
-$uri = [System.UriBuilder]$reqObj.url
-$uri.Host = "localhost"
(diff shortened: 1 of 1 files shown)


src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
-    $req.Body = $reqObj._body
-}
+$httpMethod = $requestMethods[[int]$reqObj.method]
+$uriBuilder = [System.UriBuilder]$reqObj.url
+$uriBuilder.Host = "localhost"
+$uri = $uriBuilder.ToString()
 
-if ($reqObj.headers) {
-    $req.Headers = @{}
+Write-Host $decoded
```
jhkimnew: Is this Write-Host for debugging? If so, please remove it or replace it with Write-Verbose

shhsu: Actually, not needed
Removed

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
+try {
+    $httpMsg = New-Object System.Net.Http.HttpRequestMessage -ArgumentList $httpMethod,$uri
```
jhkimnew: Put space into $httpMethod,$uri like $httpMethod, $uri

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
-try {
-    $res = Invoke-WebRequest -UseBasicParsing -UseDefaultCredentials @req
-} catch {
-    $errMsg = $_.Exception.Message
-    $res = $_.Exception.Response
-} finally {
-    if (!$res) {
-        throw $errMsg
+    if ($responseMsg.Content) {
+        $resContent = stringify $responseMsg.Content.ReadAsByteArrayAsync().GetAwaiter().GetResult()
```
jhkimnew, line 52: declaration of $resContent is missing. Please put $resContent = $null before the try statement.

src/app/resources/iis_scripts/local_http.ps1
jhkimnew: missing declaration of $httpMsg. Please put $httpMsg = $null before try statement

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
-        "body" = $content
+        "url" = $responseMsg.RequestMessage.RequestUri
+        "status" = $responseMsg.StatusCode;
+        "statusText" = $responseMsg.ReasonPhrase;
+        "type" = $responseMsg.Content.Headers.ContentType.MediaType;
+        "headers" = $responseMsg.Content.Headers;
+        "body" = $resContent
     } -Compress -Depth 100
+} finally {
+    if ($responseMsg) {
```
jhkimnew, line 64: same problem of missing declaration for $responseMsg, $clientHandler, $client and $result.

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
     foreach ($prop in $reqObj.headers | Get-Member -MemberType NoteProperty | Select-Object -ExpandProperty Name) {
-        ## NOTE: not all http headers follow the same syntax. This delimitation logic is assumed to work for all headers we use
-        $req.Headers[$prop] = $reqObj.headers.$prop -Join ","
+        $httpMsg.Headers.Add($prop, $reqObj.headers.$prop)
     }
-}
+    $clientHandler = New-Object System.Net.Http.HttpClientHandler
+    $clientHandler.UseDefaultCredentials = $true
+    $client = New-Object System.Net.Http.HttpClient -ArgumentList $clientHandler
+    $responseMsg = $client.SendAsync($httpMsg).GetAwaiter().GetResult()
```
jhkimnew, line 49: I think we should include error information in the result with adding try catch for the $client.SendAsync call so that we can show the error information to users.

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -4,70 +4,83 @@ param(
...
+## Note that index 0 maps to GET, so if $reqObj.method is null we would default to GET
 $requestMethods = @(
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Get,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Post,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Put,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Delete,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Options,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Head,
-    [Microsoft.PowerShell.Commands.WebRequestMethod]::Patch
+    [System.Net.Http.HttpMethod]::Get,
```
jhkimnew, line 20: System.Net.Http is not available in powershell unless we load the System.Net.Http.dll manually. Please add some comment from which path we should load the dll file in debugging time.

src/app/resources/iis_scripts/local_http.ps1
shhsu: done

src/app/resources/iis_scripts/local_http.ps1
shhsu: Not sure about that, in dynamic languages I don't initiate null value because it is not useful and just create clutter

src/app/resources/iis_scripts/local_http.ps1
shhsu: $client.SendAsync($httpMsg).GetAwaiter().GetResult() will throw the error from $client.SendAsync($httpMsg) and the execption will be returned from script

src/app/resources/iis_scripts/local_http.ps1
shhsu: This is not a concern because currently admin api would install .NET Core.

P.S. If admin api is not used, this script is useless anyway so right now we can assume they are always there

src/app/resources/iis_scripts/local_http.ps1
jhkimnew: Okay, I did some testing and missing declaration is not that harmful in Powershell script. We can revisit this later.

src/app/resources/iis_scripts/local_http.ps1
jhkimnew: Okay, I got it.

jhkimnew (approved):

PR #252, opened 2019-01-04
fix website routing

shhsu:
Simplify routing logic


src/app/main/app-routing.module.ts
```diff
@@ -9,9 +9,13 @@ import { ConnectComponent } from '../connect/connect.component';
 import { GetComponent } from './get.component';
 import { SettingsModule } from '../settings/settings.module'
```
jhkimnew, line 10: These three modules seem not be used. Can we remove them? 

import { SettingsModule } from '../settings/settings.module'
import { WebServerModule } from '../webserver/webserver.module'
import { WACModule } from "runtime/wac/components/wac.module";

jhkimnew: Ah, these are actually being used. I was confused because VS CODE shows them in gray color.

jhkimnew (approved):

src/app/main/app-routing.module.ts
shhsu: Correct, somehow VSCode was unable to identify that these classes are being used

Issue #222, opened 2018-11-05
No tab for logon methods of virtual directory 


jhkimnew:
https://user-images.githubusercontent.com/38521133/44787707-7975e580-abb5-11e8-8283-1e6a053088bd.png

jhkimnew:
This issue happens because enum selector has no input element.

```
@Component({
    selector: 'enum',
    template: `
        <ul [attr.disabled]="disabled ? true : null">
            <li *ngFor="let field of fields" [class.background-active]="'' + model == field.value" [attr.title]="field.title || null" [hidden]="field.hidden" (click)="select(field)">
                {{field.name}}
            </li>
        </ul>
    `,
    styles: [`
        ul {
            padding: 0;
            margin: 0;
        }

        li {
            float: left;
            cursor: pointer;
            padding: 7px;
            margin-right: 5px;
            text-align: center;
            min-width: 45px;
        }

        li:last-of-type {
            margin-right: 0;
        }
    `]
})
```

jhkimnew:
Fixed

PR #255, opened 2019-01-08
Imporve install page to allow install from path, plus some logging improvements

shhsu:


src/app/runtime/wac/components/install.component.ts
```diff
@@ -70,23 +77,33 @@ import { WACRuntime } from "runtime/runtime.wac";
 export class InstallComponent {
     private _inProgress: boolean
     private _targetHost: string
-    private _adminApiLocation: string
+    private _adminAPILocation: string
+    private _dotnetCoreLocation: string
+    invalid: boolean
```
jhkimnew, line 82: please rename the "invalid" variable with some meaningful name for readability.

src/app/runtime/wac/components/install.component.ts
```diff
@@ -70,23 +77,33 @@ import { WACRuntime } from "runtime/runtime.wac";
...
         private router: Router,
         private appContext: AppContextService,
         @Inject("Runtime") private runtime: WACRuntime,
     ) {
-        this._adminApiLocation = SETTINGS.api_download_url
+        this._adminAPILocation = SETTINGS.api_download_url
         this._targetHost = this.appContext.activeConnection.nodeName
     }
 
+    public verifyLocation(value: string, allowEmpty: boolean) {
```
jhkimnew, line 93: please rename the "value" variable with some meaningful name for readability such as setupUrl.

src/app/diagnostics/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+            case LogLevel.INFO:
+                console.info(formattedMsg)
+                break
+
+            case LogLevel.WARN:
+                console.warn(formattedMsg)
+                break
+
+            case LogLevel.ERROR:
+                console.error(formattedMsg)
```
jhkimnew, line 35: Can we set different color for error like "console.log(formattedMsg, 'background: #222; color: #bada55');"?

FYI, https://stackoverflow.com/questions/7505623/colors-in-javascript-console

jhkimnew (approved):
I added some suggestions. Please check them and update if you agree.

src/app/runtime/wac/components/install.component.ts
shhsu: done

src/app/runtime/wac/components/install.component.ts
shhsu: done

src/app/diagnostics/logger.ts
shhsu: This sounds like a good idea. I would like to put more thought into the color though. Since it custom color is a extra feature, I would like to delay the changes for now. Let's talk tomorrow when you are in and we will create a new PR

PR #257, opened 2019-01-09
Disallow installation from shared network drive

shhsu:
Install from network drive is block by WAC team's work on CredSSP feature

### DIFF
--- a/src/app/runtime/wac/components/install.component.ts
+++ b/src/app/runtime/wac/components/install.component.ts
@@ -15,15 +15,20 @@ const windowsPathValidationRegex = new RegExp('^(?:[a-z]:|\\\\\\\\[a-z0-9_.$●-
             <div class="validation-container">
                 <p>
                     <label>IIS Administration API installation location</label>
-                    <input class="form-control" type="text" (ngModelChange)="verifyLocation($event, false)" [(ngModel)]="_adminAPILocation" />
+                    <input class="form-control" type="text" [(ngModel)]="_adminAPILocation" />
                 </p>
                 <p>
-                    <label>.NET Core Runtime install location(Optional)</label>
-                    <input class="form-control" type="text" placeholder="IIS Admin API installer will fetch required .NET Core Framework online" (ngModelChange)="verifyLocation($event, true)" [(ngModel)]="_dotnetCoreLocation" />
+                    <label>.NET Core Runtime install location (Optional)</label>
+                    <input class="form-control" type="text" placeholder="IIS Admin API installer will fetch required .NET Core Framework online" [(ngModel)]="_dotnetCoreLocation" />
+                </p>
+            </div>
+            <div *ngIf='userInputError'>
+                <p class="color-error">
+                    {{userInputError}}
                 </p>
             </div>
             <p>
-                <a class="bttn background-active" [attr.disabled]="invalid ? true : null" (click)="install()">Install on {{_targetHost}}</a>
+                <a class="bttn background-active" (click)="install()">Install on {{_targetHost}}</a>
             </p>
         </div>
         <div *ngIf='_inProgress'>
@@ -79,7 +84,7 @@ export class InstallComponent {
     private _targetHost: string
     private _adminAPILocation: string
     private _dotnetCoreLocation: string
-    invalid: boolean
+    userInputError: string
 
     constructor(
         private router: Router,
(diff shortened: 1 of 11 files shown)


src/app/runtime/wac/components/install.component.ts
```diff
@@ -90,11 +95,33 @@ export class InstallComponent {
...
+        if (location && location.startsWith(`\\\\`)) {
+            return `Currently installation from shared drive is not supported, the issue is being tracked on https://github.com/Microsoft/IIS.WebManager/issues/239`
+        }
+
+        return null
     }
 
     public install() {
+        if (this.userInputError = this.verifyLocation(`IIS Administration API installation location`, this._adminAPILocation, false)) {
+            return
```
jhkimnew: put some console log or error message dialog to inform users why this verification failed.

src/app/runtime/wac/components/install.component.ts
```diff
@@ -90,11 +95,33 @@ export class InstallComponent {
...
+        }
+        if (this.userInputError = this.verifyLocation(`.NET Core Runtime install location`, this._dotnetCoreLocation, true)) {
+            return
```
jhkimnew: put some console log or error message dialog to inform users why this verification failed.

src/app/runtime/wac/components/install.component.ts
shhsu: The error shows up on screen though. Logs should be made for more implicit issues.

src/app/runtime/wac/components/install.component.ts
shhsu: The error shows up on screen though. Logs should be made for more implicit issues.

src/app/runtime/wac/components/install.component.ts
```diff
@@ -15,15 +15,20 @@ const windowsPathValidationRegex = new RegExp('^(?:[a-z]:|\\\\\\\\[a-z0-9_.$●-
             <div class="validation-container">
                 <p>
                     <label>IIS Administration API installation location</label>
-                    <input class="form-control" type="text" (ngModelChange)="verifyLocation($event, false)" [(ngModel)]="_adminAPILocation" />
+                    <input class="form-control" type="text" [(ngModel)]="_adminAPILocation" />
```
jhkimnew: It will be nice to set red outline when userInputError is set with an error.

jhkimnew (changes requested):
When error happens, we should show the hint of the problem so that users can fix the issue.

jhkimnew (changes requested):
Found some issues while testing this installation feature. Please fix them with this PR.

1) Accessibility issue on navigating keyboard only (You should test with Edge because we use the SME accessiblity inherited by WAC framework, which does not work with Chrome at this point in time)
2) IIS Administration API can be ambiguous to users. Please use the official product name such as Microsoft IIS Administration API
3) Unmatched term for .NET Core Runtime. Please use the matched and official product name for the .NET Core Runtime

src/app/runtime/wac/components/install.component.ts
```diff
@@ -13,17 +13,22 @@ const windowsPathValidationRegex = new RegExp('^(?:[a-z]:|\\\\\\\\[a-z0-9_.$●-
...
                     <label>IIS Administration API installation location</label>
-                    <input class="form-control" type="text" (ngModelChange)="verifyLocation($event, false)" [(ngModel)]="_adminAPILocation" />
+                    <input class="form-control" type="text" [(ngModel)]="_adminAPILocation"/>
                 </p>
-                <p>
-                    <label>.NET Core Runtime install location(Optional)</label>
-                    <input class="form-control" type="text" placeholder="IIS Admin API installer will fetch required .NET Core Framework online" (ngModelChange)="verifyLocation($event, true)" [(ngModel)]="_dotnetCoreLocation" />
+                <p #dotnetPrompt>
+                    <label>.NET Core Runtime install location (Optional)</label>
+                    <input class="form-control" type="text" placeholder="IIS Admin API installer will fetch required .NET Core Framework online" [(ngModel)]="_dotnetCoreLocation" />
```
jhkimnew: Admin ==> Administration

src/app/runtime/wac/components/install.component.ts
```diff
@@ -90,22 +97,56 @@ export class InstallComponent {
...
+            && this.verifyLocationPrompt(this.dotnetPrompt, true)) {
+            this._inProgress = true
+            let args: any = {
+                command: 'install',
+                adminAPILocation: this._adminAPILocation,
+            }
+            if (this._dotnetCoreLocation) {
+                args.dotnetCoreLocation = this._dotnetCoreLocation
+            }
+            var sub = this.runtime.PrepareIISHost(args).subscribe(_ => {}, _ => {}, () => {
```
jhkimnew: Please add error handling when the provided Url returns non 200 status value. 
In that case, I think we should inform users the provided URL is not valid so that users can retry.

shhsu: Good idea, fixed

src/app/runtime/wac/components/install.component.ts
shhsu: fixed

src/app/runtime/wac/components/install.component.ts
```diff
@@ -10,20 +10,23 @@ const windowsPathValidationRegex = new RegExp('^(?:[a-z]:|\\\\\\\\[a-z0-9_.$●-
 @Component({
     template: `
     <div class="center">
```
jhkimnew: I think you should add sme-focus-zone to make the navigation with arrow keys work with the controls in this page.

shhsu: done

jhkimnew (approved):

PR #262, opened 2019-01-11
Allow user to restart IIS Administration API

shhsu:


src/app/webserver/webserver.component.ts
```diff
@@ -79,4 +89,40 @@ export class WebServerComponent {
...
+                                    reject(this.failure = `Unable to restart IIS Administration API Service`)
+                                },
+                                () => { sub.unsubscribe() },
+                            )
+                        } else {
+                            reject(this.failure = `Web Server Module cannot be initialized. IIS Administration API status: ${e.Status}`)
+                        }
+                    })
+                } else {
+                    reject(this.failure = `Unknown error has occurred when trying to initialize Web Server Module`)
```
jhkimnew: I think we should expose the detailed error information for easy troubleshoot.

src/app/webserver/webserver.component.ts
```diff
@@ -79,4 +89,40 @@ export class WebServerComponent {
...
+                                        resolve(s)
+                                    })
+                                },
+                                _ => {
+                                    reject(this.failure = `Unable to restart IIS Administration API Service`)
+                                },
+                                () => { sub.unsubscribe() },
+                            )
+                        } else {
+                            reject(this.failure = `Web Server Module cannot be initialized. IIS Administration API status: ${e.Status}`)
```
jhkimnew: Maybe, the message should be more specific like:

Web Server Module cannot be initialized. Current IIS Administration API windows service status:

jhkimnew (changes requested):
I added some suggestions. NOTE: there is an accessiblity for the dialog message. Please update to fix the issue.

src/app/webserver/webserver.component.ts
```diff
@@ -79,4 +89,40 @@ export class WebServerComponent {
...
         return this._service;
     }
+
+    get server(): Promise<WebServer> {
+        return new Promise<WebServer>((resolve, reject) => {
+            this._service.server.catch(e => {
+                if (e instanceof UnexpectedServerStatusError) {
+                    this._notifications.confirm(
+                        `Restart IIS Administration API`,
+                        `IIS Administration API is currently ${e.Status}, restart?`).then(confirmed => {
```
jhkimnew: Considering the fact that restarting service can be harmful for some situations, I think we should handle two cases.

Case 1: the service is "stopped":  => call start-service instead of restart-service. Confirm message should be like "Do you want to start the service?" 

Case 2: the service is "started" but fails to connect:  => call restart-service. Confirm message should be like "Service is currently running bug can't connect to the service. Do you want to restart the service?"

src/app/webserver/webserver.component.ts
shhsu: Added basic errors. It would still not be quite readable but user should be able to refer to the console logs for more details

src/app/webserver/webserver.component.ts
shhsu: updated

src/app/webserver/webserver.component.ts
shhsu: 1. Fixed prompt message
2. If service started and fail to connect actually the user would be redirected to install page. We can talk about how we should better handle that case but case 2 would not happen for current behavior

shhsu: I have updated the question with the second sentence "Do you want to restart service"
I think the current way suffices, if the service is started and not being able to connect, the current message does show the information. I really don't know if we want to change the content of the message that much depending of the status

src/app/webserver/webserver.component.ts
```diff
@@ -79,4 +89,40 @@ export class WebServerComponent {
...
+
+    get server(): Promise<WebServer> {
+        return new Promise<WebServer>((resolve, reject) => {
+            this._service.server.catch(e => {
+                if (e instanceof UnexpectedServerStatusError) {
+                    this._notifications.confirm(
+                        `Restart IIS Administration API`,
+                        `IIS Administration API is currently ${e.Status}. Do you want to start the service?`).then(confirmed => {
+                        if (confirmed) {
+                            var sub = this._runtime.RestartIISAdministration().subscribe(
```
jhkimnew: Restarting IISAdmin API service can be dangerous because multiple users use the same service.

jhkimnew (changes requested):
Please change the design to remove unnecessary Restart service.

jhkimnew (approved):

PR #264, opened 2019-01-12
Move content header to proper place

shhsu:
This PR is currently not mergeable, I will have to rebase once https://github.com/Microsoft/IIS.WebManager/pull/262 is merged

### DIFF
--- a/src/app/resources/iis_scripts/local_http.ps1
+++ b/src/app/resources/iis_scripts/local_http.ps1
@@ -74,6 +74,7 @@ try {
                     $to = [int]::Parse($tokens[1].Trim())
                     $length = [int]::Parse($tokens[2].Trim())
                     $httpMsg.Content.Headers.ContentRange = New-Object System.Net.Http.Headers.ContentRangeHeaderValue -ArgumentList $from, $to,  $length
+                    $headerFixed = $true
                 }
                 ## possibly add more misplaced headers
             }
--- a/src/app/resources/iis_scripts/local_http.ps1
+++ b/src/app/resources/iis_scripts/local_http.ps1
@@ -1,3 +1,4 @@
+[CmdletBinding()]
 param(
     [string]
     $requestBase64
@@ -6,6 +7,23 @@ param(
 $ErrorActionPreference = "Stop"
 Add-Type -AssemblyName System.Net.Http
 $contentEncoding = [System.Text.Encoding]::UTF8
+
+######################## Utilities ###############################
+$verbose = $PSBoundParameters['verbose']
+
+if ($verbose) {
+    $logDir = Join-Path $env:UserProfile 'wac-iis-logs'
+    if (!(Test-Path $logDir)) {
+        mkdir $logDir
+    }
+    $logFile = Join-Path $logDir 'local_http.log'
+}
+
+function LogVerbose([string] $msg) {
+    if ($verbose) {
+        Add-Content -Value $msg -Path $logFile -Force
+    }
+}
 
 function stringify([System.Byte[]]$content) {
     if (!$content) {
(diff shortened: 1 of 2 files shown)


src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -36,12 +56,33 @@ $uri = $uriBuilder.ToString()
...
+            if ($httpMsg.Content) {
+                if ($prop -like "content-type") {
+                    $httpMsg.Content.Headers.ContentType = New-Object System.Net.Http.Headers.MediaTypeHeaderValue $headerValue
+                    $headerFixed = $true
+                } elseif ($prop -like "content-range") {
+                    $tokens = $prop.Split("/-")
+                    $from = [int]::Parse(($tokens[0].Trim() -split " ")[-1])
+                    $to = [int]::Parse($tokens[1].Trim())
+                    $length = [int]::Parse($tokens[2].Trim())
+                    $httpMsg.Content.Headers.ContentRange = New-Object System.Net.Http.Headers.ContentRangeHeaderValue -ArgumentList $from, $to,  $length
```
jhkimnew, line 76: I think you missed "$headerFixed = $true" for this block

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -36,12 +56,33 @@ $uri = $uriBuilder.ToString()
...
+                }
+                ## possibly add more misplaced headers
+            }
+            if ($headerFixed) {
+                LogVerbose "Replaced header ${prop}: ${headerValue}"
+            } else {
+                LogVerbose "Failed to add header ${prop}: ${headerValue}"
```
jhkimnew: Can we throw exception here for easy debugging? 
If there is no exception, it won't be easy what headers were not failed to added.

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -36,12 +56,33 @@ $uri = $uriBuilder.ToString()
...
     foreach ($prop in $reqObj.headers | Get-Member -MemberType NoteProperty | Select-Object -ExpandProperty Name) {
-        $httpMsg.Headers.Add($prop, $reqObj.headers.$prop)
+        $headerValue = $reqObj.headers.$prop
+        if (!$httpMsg.Headers.TryAddWithoutValidation($prop, $headerValue)) {
+            $headerFixed = $false
+            if ($httpMsg.Content) {
+                if ($prop -like "content-type") {
+                    $httpMsg.Content.Headers.ContentType = New-Object System.Net.Http.Headers.MediaTypeHeaderValue $headerValue
+                    $headerFixed = $true
+                } elseif ($prop -like "content-range") {
```
jhkimnew, line 71: Would you add some comment why content-range is handled here?
Usually content-range is for response headers, not for request headers.

See https://stackoverflow.com/questions/716680/difference-between-content-range-and-range-headers more information.

src/app/resources/iis_scripts/local_http.ps1
```diff
@@ -36,12 +56,33 @@ $uri = $uriBuilder.ToString()
...
+                    $length = [int]::Parse($tokens[2].Trim())
+                    $httpMsg.Content.Headers.ContentRange = New-Object System.Net.Http.Headers.ContentRangeHeaderValue -ArgumentList $from, $to,  $length
+                }
+                ## possibly add more misplaced headers
+            }
+            if ($headerFixed) {
+                LogVerbose "Replaced header ${prop}: ${headerValue}"
+            } else {
+                LogVerbose "Failed to add header ${prop}: ${headerValue}"
+            }
```
jhkimnew, line 85: I think we should also throw exception here to make it easy to find the unfixed header issue.

jhkimnew (approved):
Approving with below suggestion:

It seems that Content-Range request header was used for Angular bug and it seems that the bug is fixed already. So, please create a new work-item to clean up adding the Content-Range header and update this as well because Content-Range is not expected in request headers.

PR #282, opened 2019-01-16
Allow different styles for wac

shhsu:


src/gulpfile.js
```diff
@@ -73,8 +73,18 @@ gulp.task('generate-resjson', (cb) => {
     runSequence(['generate-resjson-json', 'generate-resjson-interface'], 'merge-localized-json', cb);
 });
 
+gulp.task('generate-angular-cli-json', (_) => {
```
jhkimnew: Please move this new task definition above the the task definition of generate-powershell to match the order of runSequence(['generate-angular-cli-json', 'generate-powershell', 'generate-svg', 'generate-resjson'], cb).

src/themes/themes.wac.css
```diff
@@ -0,0 +1,256 @@
+﻿.red {
```
jhkimnew, line 1: Please put some comment why and how you created this style and the difference from the original style.

src/gulpfile.js
shhsu: done

src/themes/themes.wac.css
shhsu: Not sure about this. It's a copy of theme.blue.css but that's not a relevant information from this point onwards and the difference will only grow.

src/angular-cli.site.json
```diff
@@ -0,0 +1,9 @@
+{
```
jhkimnew, line 1: 'site' is ambiguous. Can we rename with angular-cli.default.json or something like that?

shhsu: 'default' is too generic. There are 2 scenarios of this repo: wac extension and standalone website. Hence I call the scenario site

src/gulpfile.js
```diff
@@ -73,8 +73,18 @@ gulp.task('generate-resjson', (cb) => {
     runSequence(['generate-resjson-json', 'generate-resjson-interface'], 'merge-localized-json', cb);
 });
 
+gulp.task('generate-angular-cli-json', (_) => {
+    var override = process.argv.slice(3).find(function(s, _, __) { return s.startsWith('--env=') })
+    var scenario = override ? override.split('=').pop().split('.')[0] : 'site'
+    return gulp.src(['angular-cli.template.json', `angular-cli.${scenario}.json`])
```
jhkimnew: put some comment here for readability.

jhkimnew (approved):
Approving with some suggestions. Please check them and update if you agree.

PR #286, opened 2019-01-18
Install fix ups: minimal permissions, install button fix

shhsu:


jhkimnew (approved):
Put some comments for the complicated logic you updated here in order to enhance the readability in the future.

PR #289, opened 2019-01-19
Removed hard-coded nuget package version

jhkimnew:
Removed the hard-coded version added a new --version parameter for the build.ps1.

Ex:
    --version=0.1.$(Build.BuildNumber)

This will be converted like 0.1.555 and the nuget package will be published @https://www.myget.org/feed/iis-wac/package/nuget/microsoft.iis-wac-extension with the specified version.

In manual build, we can also set the parameter like this as well.

    --version=1.0.0-RC

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -26,16 +26,12 @@ function ShouldNPMInstall {
 
 $purge = $args | Where-Object { $_ -like "--purge" }
 $pack = $args | Where-Object { $_ -like "--pack" }
-$pack_build = $args | Where-Object { $_.startsWith("--pack.build=") }
-if ($pack_build) {
-    # Example: --pack.build=0.1.$(Build.BuildNumber)/$(Build.SourceVersion)
-    Write-Host ($pack_build)
-    $tokens = $pack_build.split("=")[1].trim().split("/")  
-    $pack_build = $pack_build = $tokens[0]
-} else {
-    $pack_build = "0.1.0" # default value
+$version = $args | Where-Object { $_.startsWith("--version=") }
+if ($version) {
+    # Example: --version=0.1.$(Build.BuildNumber)
+    Write-Host ($version)
+    $version = $version.split("=")[1]
 }
-Write-Host ("pack_build: " + $pack_build)
 
 if (!(Get-Command "npm" -ErrorAction SilentlyContinue)) {
     throw "npm is required in PATH"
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -26,6 +26,16 @@ function ShouldNPMInstall {
 
 $purge = $args | Where-Object { $_ -like "--purge" }
 $pack = $args | Where-Object { $_ -like "--pack" }
+$pack_build = $args | Where-Object { $_.startsWith("--pack.build=") }
```
shhsu: this variable name is a bit verbose. Use --version instead. Note that you can argue the version number is not just for nuget pack and it is possible that we could pass the argument down in the build process and mark the manifest.json too

build.ps1
```diff
@@ -26,6 +26,16 @@ function ShouldNPMInstall {
 
 $purge = $args | Where-Object { $_ -like "--purge" }
 $pack = $args | Where-Object { $_ -like "--pack" }
+$pack_build = $args | Where-Object { $_.startsWith("--pack.build=") }
+if ($pack_build) {
+    # Example: --pack.build=0.1.$(Build.BuildNumber)/$(Build.SourceVersion)
+    Write-Host ($pack_build)
+    $tokens = $pack_build.split("=")[1].trim().split("/")
```
shhsu: ``-Split`` should suffice because we are passing the string in the command line below and spaces would be ignored

build.ps1
```diff
@@ -26,6 +26,16 @@ function ShouldNPMInstall {
...
 $purge = $args | Where-Object { $_ -like "--purge" }
 $pack = $args | Where-Object { $_ -like "--pack" }
+$pack_build = $args | Where-Object { $_.startsWith("--pack.build=") }
+if ($pack_build) {
+    # Example: --pack.build=0.1.$(Build.BuildNumber)/$(Build.SourceVersion)
+    Write-Host ($pack_build)
+    $tokens = $pack_build.split("=")[1].trim().split("/")  
+    $pack_build = $pack_build = $tokens[0]
+} else {
+    $pack_build = "0.1.0" # default value
```
shhsu: I am thinking we don't want to give a default version here just in case everyone who missed this parameter would  push to the same spot. What happens if --version is missing to nuget push? We should match that behavior

shhsu (changes requested):

build.ps1
jhkimnew: Done. Removed the default valu.e

FYI,
Without the -version, it fails with the below error. 

```
C:\gitroot\IIS.WebManager\src>nuget pack . -OutputDirectory ..\dist
Attempting to build package from 'wac-iis.nuspec'.
Value cannot be null or an empty string.
Parameter name: value
```

build.ps1
jhkimnew: done

build.ps1
jhkimnew: done

shhsu (approved):

PR #294, opened 2019-01-31
take away cors settings

shhsu:


jhkimnew:
@shhsu : Did you get confirmation from @shirhatti about this change?

There will be a new usability issue that users can't use the manage.iis.net because the default cors setting which allows the origin of https://manage.iis.net got removed by this change.

shirhatti:
@shhsu Just curious what is the impact of install time from these changes?

shhsu:
It's neglectable since we were already editing the file on install anyway. @shirhatti

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -212,73 +218,74 @@ if (!$config.security.users.owners.Contains($user)) {
...
-    if ($saveConfig) {
-        LogVerbose "Saving config..."
-        if ($devMode) {
-            throw "Cannot edit config file in dev mode, please add $user to ""$iisAdminOwners"" group manually"
+if ($saveConfig) {
+    LogVerbose "Saving config..."
+    if ($devMode) {
+        throw "Cannot edit config file in dev mode, please add $user to ""$iisAdminOwners"" group manually"
+    }
+    if (!(Get-LocalGroupMember -Group "Administrators" -Member $user -ErrorAction SilentlyContinue)) {
```
jhkimnew, line 229: One globalization issue here. We can't use the hard-coded name for "Administrators" local group. In Germany, it is a different one. Please find out how to solve the issue. FYI, I found one article about that. https://stackoverflow.com/questions/3212862/how-can-i-get-the-local-group-name-for-guests-administrators

src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -212,73 +218,74 @@ if (!$config.security.users.owners.Contains($user)) {
...
+        $dirAcl.RemoveAccessRule($access) | Out-Null
+        $updateAcl = $true
+    }
+    if ($updateAcl) {
+        LogVerbose "Updating Acl to remove readonly access..."
+        Set-Acl -Path $apiHome -AclObject $dirAcl | Out-Null
+    }
+    $modifyAcess = [System.Security.AccessControl.FileSystemRights]::Modify
+    $allow = [System.Security.AccessControl.AccessControlType]::Allow
+    $dirAccessGranted = $dirAcl.Access | Where-Object { ($_.IdentityReference.Value -eq "BUILTIN\Administrators") -and ($_.AccessControlType -eq $allow) -and (($_.FileSystemRights -bAnd $modifyAcess) -eq $modifyAcess) }
```
jhkimnew, line 245: One more. Please check if all the hard-coded name (BUILTIN\Administrators) works in Germany OS or not.

src/app/resources/iis_scripts/admin_api_util.ps1
jhkimnew: One more article. 
https://stackoverflow.com/questions/24426405/find-admin-group-name

shhsu: That's a good point. But this is a different work item. Can you open a new issue and I will fix it. This PR is just for CORS

jhkimnew (approved):
We decided to fix the globalization issue as another work item.

PR #295, opened 2019-01-31
Fixing incorrectly added _pingPopup.clos()

jhkimnew:
I misunderstood the logic. We don't need to call _pingPopup.close() because it is supposed to be done by other function such as this.reset().

### DIFF
--- a/src/app/connect/connect.service.ts
+++ b/src/app/connect/connect.service.ts
@@ -66,11 +66,10 @@ export class ConnectService {
 
             // Raw request to url causes certificate acceptance prompt on IE.
             this._client.raw(conn, "/api")
-            .subscribe(
-                _ => {}, 
-                e => {}, // Ignore errors
-                () => { this._pingPopup.close(); } 
-            );
+                .subscribe(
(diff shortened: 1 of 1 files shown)


shhsu (approved):
Please make sure the commit is named to reflect what this change it. It's not a simple rollback of the entire past commit. Just a fix

jhkimnew:
Added some information from Jimmyca15 
```
The popup allows users to receive the browser prompt to allow requests to the IIS Administration API which uses an untrusted certificate. The certificate may be untrusted when accessing from a computer other than the one that the API was installed on. AJAX requests to an untrusted certificate are simply blocked, but opening up the popup allows the user to bypass it with the in-browser prompt.

The tab shouldn’t be closed before successfully pinging the API, otherwise it will just be closed before the user can click through the untrusted certificate prompt.

One option is to update the IIS Administration API so that the page that is loaded in the popup automatically closes itself. Or we could display a prompt to the user

“Hello, this page is used to ensure that a connection can be established. Everything is in order. Please “click here” to close!”
```

PR #304, opened 2019-02-06
Jhkim/fix case sensitive issue

jhkimnew:
Get-LocalGroupMember works case-sensitively and fails in localized languages such as Germany.
As a work-around, we decided to use the where-object which works case-insensitively.

### DIFF
--- a/src/app/resources/iis_scripts/admin_api_util.ps1
+++ b/src/app/resources/iis_scripts/admin_api_util.ps1
@@ -123,7 +123,7 @@ function EnsureGroup($group) {
 }
 
 function EnsureMember($group, $userOrGroup) {
-    $modify = !(Get-LocalGroupMember -Group $group -Member $userOrGroup -ErrorAction SilentlyContinue)
+    $modify = !(Get-LocalGroupMember -Group $group | Where-Object { $_.Name -eq $user })
     if ($modify) {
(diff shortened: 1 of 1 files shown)


src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -251,7 +251,7 @@ if ($saveConfig) {
         throw "Cannot edit config file in dev mode, please add $user to ""$iisAdminOwners"" group manually"
     }
     ## Check the current user is a member of the Administrators local user group
-    if (!(Get-LocalGroupMember -SID $administratorsSID -Member $user -ErrorAction SilentlyContinue)) {
+    if (!(Get-LocalGroupMember -SID $administratorsSID -ErrorAction SilentlyContinue | Where-Object { $_.Name -eq $user } )) {
```
shhsu, line 256: are you sure -eq works? I thought only -like would work

shhsu (changes requested):

src/app/resources/iis_scripts/admin_api_util.ps1
shhsu: Also, since it is a non-trivial workaround, a comment is required

jhkimnew: yes, the -eq works.
I added some comments.

shhsu (approved):

PR #313, opened 2019-02-08
Service name update

shhsu:


jhkimnew (approved):
Thanks for the fix. 👍

PR #317, opened 2019-02-11
Fix nuget package version

shhsu:


shirhatti:
This won't work? You're publishing both branches to the same feed. You'll just keep clobbering the package

shhsu:
@shirhatti 

Currently I am using branch name to differentiate the results of different branches. Note that the package version is the build name and the build name includes branch

jhkimnew (approved):

PR #320, opened 2019-02-13
nuspec fixes

shhsu:


shirhatti:
I'd hold off on merging this just yet. Changing the package id will make us lose the notification that WAC gives you when an upgrade is available.

Based on how many folks end-up using the preview, I want to see how many will be hurting by the missing upgrade feature and hence will like to wait and see.

jhkimnew:
@shhsu Would you remove the change of the package id from this PR so that we can fix the other issues first?

shhsu:
we will update package name later

PR #324, opened 2019-02-14
Assign file acl to user instead of local administrator

shhsu:
Originally we assign modify acl to local administrator and it persists. However since we decide to rollback any acl changes after we edit user group, we can assign the acl directly to user

### DIFF
--- a/src/app/resources/iis_scripts/admin_api_util.ps1
+++ b/src/app/resources/iis_scripts/admin_api_util.ps1
@@ -143,30 +143,6 @@ function GetIISAdminHome($procs) {
     }
 }
 
-function ConvertTo-NTAccount($From)
-{
-    if ($From -is [System.Security.Principal.NTAccount]) {
-        return $From
-    }
-    if ($From -is [System.Security.Principal.SecurityIdentifier]) {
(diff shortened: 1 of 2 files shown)


src/app/resources/iis_scripts/admin_api_util.ps1
```diff
@@ -245,34 +246,17 @@ if (!$config.security.users.owners.Contains($user)) {
...
-        $updateAcl = $true
-    }
-    if ($updateAcl) {
-        LogVerbose "Updating Acl to remove readonly access..."
-        Set-Acl -Path $apiHome -AclObject $dirAcl | Out-Null
-    }
-
     $modifyAcess = [System.Security.AccessControl.FileSystemRights]::Modify
     $allow = [System.Security.AccessControl.AccessControlType]::Allow
-    $builtInAdministrators = (ConvertTo-NTAccount $administratorsSID).value
```
jhkimnew, line 273: The function ConvertTo-NTAccount() is not used anymore. Please remove the function because we don't need that.

jhkimnew (approved):
Approved. Please delete ConvertTo-NTAccount() before merging this change.

PR #328, opened 2019-02-20
Fixed case-sensitive paramer value from ALL to all

jhkimnew:
The value of --output-hashing is case sensitive. We have to use "all" not "ALL".

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -57,7 +57,7 @@ if ($pack) {
     if (
         ($outputHashingIndex -eq -1) -Or
         ($outputHashingIndex+1 -ge $args.Count) -Or
-        ($args[$outputHashingIndex+1].Equals("all"))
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -59,7 +59,7 @@ if ($pack) {
         ($outputHashingIndex+1 -ge $args.Count) -Or
         ($args[$outputHashingIndex+1].ToLower() -ne "all")
```
shhsu: can you remove this .ToLower() as well

jhkimnew: @shhsu done.

shhsu: actually, I just realized, `-ne` does not work here at all because it is not case sensitive, can you use `.equals()` instead

jhkimnew: Fixed.

shhsu (approved):

PR #335, opened 2019-03-27
Upgrade dependencies: Angular, Gulp, Node, WAC, etc...

shhsu:


src/angular.site.json
```diff
@@ -0,0 +1,44 @@
...
+  "projects": {
+    "microsoft.iis.web-manager": {
+      "architect": {
+        "build": {
+          "configurations": {
+            "wac": {
+              "fileReplacements": [
+                {
+                  "replace": "app/environments/environment.ts",
+                  "with": "app/environments/environment.prod.ts"
```
jhkimnew, line 11: There seems to be one garbage file named environemnt.prod.ts. Please remove the file.

```
C:\gitroot\IIS.WebManager\src>dir app\environments
...
03/28/2019  09:51 AM                78 environemnt.prod.ts
03/28/2019  10:13 AM               543 environment.prod.ts
...

```

shhsu: Fixed

src/app/common/httpclient.ts
```diff
@@ -83,7 +80,18 @@ export class HttpClient {
...
+        return new Promise((resolve, reject) => {
+            let sub: Subscription = this.requestOnConnection(url, options, warn).subscribe(
+                v => resolve(v),
+                e => {
+                    let err: any = this.runtime.HandleConnectError(e);
+                    if (err) {
+                        reject(err);
+                    }
+                },
+                () => sub.unsubscribe(),
```
jhkimnew: The variable "sub" here is local variable. So, this shouldn't be used inside of the complete callback function because these are called asynchronously. 

Please see this article for the detail.

https://www.pluralsight.com/guides/javascript-callbacks-variable-scope-problem

If we really want to unsubscribe, I think the variable should be saved in the parent scope.

src/app/runtime/runtime.wac.ts
```diff
@@ -54,85 +50,40 @@ export class WACRuntime implements Runtime {
...
-    public DestroyContext() {
+    public OnAppDestroy(): void {
         if (this._tokenId) {
-            this.powershellService.run(PowerShellScripts.token_utils, {
+            this.powershellService.run(PowerShellScripts.token_utils.script, {
                 command: 'delete',
                 tokenId: this._tokenId,
                 apiHost: this._apiHost,
-            }).finally(() => this.appContext.ngDestroy()).subscribe()
+            }).subscribe(null, null, () => this.appContext.ngDestroy());
```
jhkimnew: Considering this is a new code, please check that if we need to add unsubscribe(), which is missing here and add it if required.

src/app/versioning/version.service.ts
```diff
@@ -40,7 +35,7 @@ export class VersionService {
                 this.checkVersions();
             }
 
-            IntervalObservable.create(interval).subscribe(_ => {
+            interval(delay).subscribe(_ => {
```
jhkimnew, line 38: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/versioning/version.service.ts
```diff
@@ -78,7 +73,7 @@ export class VersionService {
 
     private notificationExists(): boolean {
         let notifications: Notification[];
-        let sub = this._notificationService.notifications.first().subscribe(n => {
+        let sub = this._notificationService.notifications.pipe(first()).subscribe(n => {
```
jhkimnew, line 76: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/versioning/version.service.ts
```diff
@@ -138,7 +133,7 @@ export class VersionService {
 
     private connected() {
         let connection = null;
-        this._connectSvc.active.first().subscribe(c => {
+        this._connectSvc.active.pipe(first()).subscribe(c => {
```
jhkimnew, line 136: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -96,7 +89,7 @@ export class AppPoolsService {
             //
             // Ping
             if (pool.status == Status.Starting) {
-                let ob = IntervalObservable.create(1000).subscribe(i => {
+                let ob = interval(1000).subscribe(i => {
```
jhkimnew: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -118,7 +111,7 @@ export class AppPoolsService {
             //
             // Ping
             if (pool.status == Status.Stopping) {
-                let ob = IntervalObservable.create(1000).subscribe(i => {
+                let ob = interval(1000).subscribe(i => {
```
jhkimnew: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -140,7 +133,7 @@ export class AppPoolsService {
         this.stop(pool);
 
         if (pool.status == Status.Stopping) {
-            let ob = IntervalObservable.create(200).subscribe(i => {
+            let ob = interval(200).subscribe(i => {
```
jhkimnew: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/webserver/files/webfile-list.ts
```diff
@@ -308,7 +305,7 @@ export class WebFileListComponent implements OnInit, OnDestroy {
     }
 
     private filter() {
-        this._svc.files.take(1).subscribe(files => {
+        this._svc.files.pipe(take(1)).subscribe(files => {
```
jhkimnew, line 308: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/webserver/websites/binding-list.component.ts
```diff
@@ -283,7 +280,7 @@ export class BindingItem implements OnInit, OnChanges {
             this.model.protocol = "https";
 
             if (!this.model.certificate) {
-                this._certSelect.changes.first().subscribe(c => {
+                this._certSelect.changes.pipe(first()).subscribe(c => {
```
jhkimnew, line 283: This looks like a legacy code. For safety, please check if we need to call unsubscribe or not.

src/app/common/httpclient.ts
jhkimnew: Even though let has some magic to make the variable in the upper layer, I don't think it is a good practice to save the subscription object variable inside a callback function and make it not readable.

jhkimnew (changes requested):
I found that we don't call unsubscribe() after calling subscribe() for some cases. Please investigate those and fix them if required. 

And please refactor the code to save the subscription object as the other places or the code example of rxjs reference.

src/app/common/httpclient.ts
shhsu: This is not the only place where this kind of logic exists. https://github.com/Microsoft/IIS.WebManager/blob/dev/src/app/webserver/webserver.service.ts#L67
I thought we have agreed offline to address this kind of coding pattern issue later?
I disagree with the readability though. The variable is used after it was declared. What is unreadable about this?

src/app/runtime/runtime.wac.ts
shhsu: Note that this code is OnAppDestroy. All callbacks would go out of scope right after this method. We shouldn't need to do it. But I will add it just to be consistent

shhsu:
I cannot do the refactoring in this PR. Please create a new issue if you want some investigation on this unsubscribe issue

src/app/common/httpclient.ts
jhkimnew: Yeah, I agreed about that yesterday but I thought we have still the time to fix the issue instead of postponing. But I am okay to investigate this issue in a separate work item.
I opened this issue and assigned it to you. 
https://github.com/Microsoft/IIS.WebManager/issues/336

In my experience, using local variable in callback function is problematic but should be avoided. But if that was used, I have to figure out why that was used and that makes me to read the code difficult. Maybe we can discuss about this in the team meeting. :-)

jhkimnew (approved):
I found one issue #336 but we decided to postpone the investigation about the issue.

src/app/common/httpclient.ts
shhsu: Removed unsubscribe method

PR #346, opened 2019-04-19
Pin context related tabs in vtab

shhsu:


jhkimnew:
Vertical line of apppool is too short because the length of line is determined by the number of modules. 
I think we should add a vertical divider to fix this issue.

jhkimnew:
We decided not to show the initial root bread crumb (Microsoft IIS).

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+                <ng-container *ngIf="!(module.routerLink)">
+                    <dynamic [name]="module.component_name" [module]="module" [data]="module.data"></dynamic>
+                </ng-container>
+            </item>
+        </ng-container>
+    </vtabs>
+</div>
+`,
+    styles: [`
+    :host >>> .sidebar > vtabs .vtabs > .items {
```
jhkimnew, line 61: ">>>"  seems to be deprecated according to https://angular.io/guide/component-styles. Please update this not to use the deprecated styles.

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const CONTEXT_MODULES = [
+    new RouteReference(WebServerModuleName, WebServerModuleIcon, [`/webserver/${SectionHelper.normalize(WebServerModuleName)}+general`]),
+    new RouteReference(WebSitesModuleName, "fa fa-globe", [`/webserver/${SectionHelper.normalize(WebSitesModuleName)}`]),
+    new RouteReference(AppPoolsModuleName, "fa fa-cogs", [`/webserver/${SectionHelper.normalize(AppPoolsModuleName)}`]),
+    new RouteReference(FileSystemModuleName, "fa fa-files-o", [`/webserver/${SectionHelper.normalize(FileSystemModuleName)}`]),
+];
+
+export class Feature extends ComponentReference {
+    data: any;
```
jhkimnew, line 22: Please avoid using any if we can define type. Or, please explain what is supposed to be set with the data.

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+export interface FeatureContext {
+    _links: any
```
jhkimnew, line 26: Please avoid using any if we can define type. Or, please explain what is supposed to be set with _links.

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    @Input() generalTabName: string = "General";
+    @Input() generalTabIcon: string = "fa fa-wrench";
+    @Input() generalTabCategory: string;
+    @Input() model: FeatureContext;
+    @Input() resource: string;
+    @Input() subcategory: string;
+    @Input() include: GlobalModuleReference[] = [];
+    @Input() promote: string[] = [];
+    @Input() contexts: any[] = [];
+    features: Feature[];
```
jhkimnew, line 79: Do we need to initialize Feature[] with adding "= []"?

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+
+@Component({
+    selector: 'feature-vtabs',
+    template: `
+<div class="sidebar crumb" [class.nav]="IsActive">
+    <vtabs [markLocation]="true" (activate)="Refresh()" [defaultTab]="default" [categories]="['${HomeCategory}', subcategory]">
+        <item [name]="generalTabName" [ico]="generalTabIcon" [category]="generalTabCategory || subcategory">
+            <ng-content select=".general-tab"></ng-content>
+        </item>
+        <item *ngFor="let module of features" [name]="module.name" [ico]="module.ico" [category]="subcategory">
```
jhkimnew, line 47: please rename with feature.
let feature of features

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    @Input() generalTabCategory: string;
+    @Input() model: FeatureContext;
+    @Input() resource: string;
+    @Input() subcategory: string;
+    @Input() include: GlobalModuleReference[] = [];
+    @Input() promote: string[] = [];
+    @Input() contexts: any[] = [];
+    features: Feature[];
+
+    @ViewChild(forwardRef(() => VTabsComponent)) vtabs: VTabsComponent;
```
jhkimnew, line 81: Please put some comment what "forwardRef(() => VTabsComponent)" means.

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    :host >>> .sidebar > vtabs .vtabs > .content {
+        top: 96px;
+    }
+`],
+}) export class FeatureVTabsComponent implements OnInit, AfterViewInit {
+    @Input() default: string;
+    @Input() generalTabName: string = "General";
+    @Input() generalTabIcon: string = "fa fa-wrench";
+    @Input() generalTabCategory: string;
+    @Input() model: FeatureContext;
```
jhkimnew, line 73: Put some comment for supported value for model such as "pool", "app", "site" for readerbility.

src/app/common/feature-vtabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    Refresh() {
+        this.options.refresh();
+    }
+
+    ngOnInit(): void {
+        const apiNames = Object.keys(this.model._links);
+        const featureSet: Feature[] = [];
+        let promoted = new Map<string, Feature>();
```
jhkimnew, line 102: rename all the variable names such as promoted, include and includes. 
Maybe we should simply add "childModules" before or after the existing variable names. 

FYI, the variable name include  can be confused with the "includes" function.

src/app/common/feature-vtabs.component.ts
shhsu: I can't, removing it seem to move everything out of this section

src/app/common/feature-vtabs.component.ts
shhsu: Same as above, this is not designed by me. This item is meant to be dynamic.

src/app/common/feature-vtabs.component.ts
shhsu: I will write a more general documentation. Check the next revision of this PR

jhkimnew (approved):
We decided to fix the issue of the short line in apppool. 

I found two accessibility issues. These should be fixed in a separated work item.

1. When feature module selected after selecting sites listview item, the module's style is not shown in focused state. 
2. Keyboard enter key does not work for web sites list view item.

shhsu:
@jhkimnew Make sure to open some issues for above comments. Thanks

PR #379, opened 2019-06-11
use either exception or error while unwraping AjaxError object

jhkimnew:
When authentication error (401) happens, exception is used instead of error.
We decided to handle both of them.

The screenshot after the update which I got while testing unauthorized users. (we can test this with updating appsettings.json and removing the current user):

image

### DIFF
--- a/src/app/runtime/wac/services/powershell-service.ts
+++ b/src/app/runtime/wac/services/powershell-service.ts
@@ -120,8 +120,13 @@ export class PowershellService {
       catchError((e, _) => {
         let rethrow = e;
         // WAC wrap the powershell error message around this AjaxError object. We would unwrap it for easier readability
-        if (e.name == "AjaxError" && e.status == 400 && e.response.error) {
-          rethrow = e.response.error;
+        if (e.name == "AjaxError" && e.status == 400) {
+          if (e.response.error && !e.response.exception) {
+            rethrow = e.response.error;
+          } 
+          if (!e.response.error && e.response.exception) {
+            rethrow = e.response.exception;
+          }
         }
         throw rethrow;
       }),
--- a/src/app/runtime/wac/services/powershell-service.ts
+++ b/src/app/runtime/wac/services/powershell-service.ts
@@ -119,7 +119,7 @@ export class PowershellService {
       }),
       catchError((e, _) => {
         let rethrow = e;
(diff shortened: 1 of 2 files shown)


src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -119,9 +119,14 @@ export class PowershellService {
...
-          rethrow = e.response.error;
+        // WAC wrap the powershell error (or exception) message around this AjaxError object. We would unwrap it for easier readability
+        if (e.name == "AjaxError" && e.status == 400) {
+          if (e.response.error && !e.response.exception) {
+            rethrow = e.response.error;
+          } 
+          if (!e.response.error && e.response.exception) {
+            rethrow = e.response.exception;
+          }
         }
```
shhsu, line 130: Looks like `error` is present if script fails with `Write-Error`. `exception` is present if script fails with `throw`.
We need to verify this assumption and always check if they both exists, output warning if that happens.
The script as is only silently override `rethrow` variable and swallow the problem

src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -119,9 +119,14 @@ export class PowershellService {
       }),
       catchError((e, _) => {
         let rethrow = e;
-        // WAC wrap the powershell error message around this AjaxError object. We would unwrap it for easier readability
-        if (e.name == "AjaxError" && e.status == 400 && e.response.error) {
-          rethrow = e.response.error;
+        // WAC wrap the powershell error (or exception) message around this AjaxError object. We would unwrap it for easier readability
+        if (e.name == "AjaxError" && e.status == 400) {
+          if (e.response.error && !e.response.exception) {
```
shhsu, line 124: Now we are catching exception, this break the logic in `runtime.wac.ts` `GetApiKey`. Update accordingly

shhsu (changes requested):

src/app/runtime/wac/services/powershell-service.ts
shhsu: Another solution would be what we talked about. If they are both present. Do not unwrap

jhkimnew: Fixed.

shhsu (approved):

PR #390, opened 2019-06-14
Remove model header

shhsu:


src/app/common/status-controller.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    <button class="stop" title="Stop" [attr.disabled]="!controller.CanStop() || null" (click)="controller.Stop()">Stop</button>
+</div>`,
+    styles: [`
+.controller {
+    padding-bottom: 15px;
+}
+`],
+})
+export class StatusControllerComponent {
+    @Input() restartLabel: string = "Restart";
```
jhkimnew: Please put comment that the example of the input values for restartLabel for better readability such as Recyle or Restart.
Otherwise, it can be confusing why only the restartLabel is declared as a input value unlike others such as stop and start.

src/app/common/status-controller.component.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.Invoke(this.RestartImpl());
+    }
+
+    Invoke(operation: Promise<any>) {
+        this.inProgress = true;
+        operation.finally(() => this.inProgress = false);
+    }
+}
+
+@Component({
```
jhkimnew, line 49: Please put @Component first before class. I think that is the coding convention.

src/app/common/status-controller.component.ts
```diff
@@ -0,0 +1,82 @@
...
+export abstract class StatusController {
+    private inProgress: boolean = false;
+    abstract StartImpl(): Promise<any>;
+    abstract StopImpl(): Promise<any>;
+    constructor(
+        private model: StatusModel,
+    ) {}
+
+    RestartImpl(): Promise<any> {
+        return this.StartImpl().then(this.StopImpl);
```
jhkimnew: Wrong order. Stop => Start, not Start => Stop

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -123,14 +123,19 @@ export class AppPoolsService {
         });
     }
 
-    recycle(pool: ApplicationPool) {
+    recycle(pool: ApplicationPool): Promise<any> {
         if (pool.status == Status.Stopped) {
             return this.start(pool);
         } else {
-            return this.stop(pool).then(_ => this.start(pool));
+            return this.restart(pool);
```
jhkimnew: The restart() of webserver.service.ts also uses the old pattern. Please update it with this new style of using await to be consistent.

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -123,26 +123,17 @@ export class AppPoolsService {
...
-            if (pool.status == Status.Stopping) {
-                let ob = interval(200).subscribe(i => {
-                    if (pool.status == Status.Stopped || i >= 450) {
-                        ob.unsubscribe();
-                        this.start(pool);
-                    }
-                });
-            }
-        });
+    async restart(pool: ApplicationPool): Promise<any> {
```
jhkimnew, line 134: In the previous code, we handles exception error and do some retry logic. But in the new code, there is no such a behavior.

jhkimnew:
@shhsu In the site level, the status does not work if the site is using an apppool which is removed.
This is not a regression issue. If you don't want to fix the issue, please create a new issue.  

image

src/app/webserver/app-pools/app-pools.service.ts
```diff
@@ -123,26 +123,17 @@ export class AppPoolsService {
...
+        await this.stop(pool)
```
jhkimnew: missing ;

jhkimnew: There is another one in D:\gitroot\IIS.WebManager\src\app\webserver\webserver.service.ts.

shhsu: fixed

src/app/common/status-controller.component.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export abstract class StatusController {
+    private inProgress: boolean = false;
+    abstract StartImpl(): Promise<any>;
+    abstract StopImpl(): Promise<any>;
+    constructor(
+        private model: StatusModel,
+    ) {}
+
+    async RestartImpl(): Promise<any> {
```
jhkimnew: please check the naming convention, pascal?

src/app/webserver/webserver-general.component.ts
```diff
@@ -1,63 +1,47 @@
...
+</fieldset>
 `,
-    styles: [`
-    .controller {
-        padding-bottom: 50px;
-    }
-`],
 })
-export class WebServerGeneralComponent {
+export class WebServerGeneralComponent implements OnInit {
```
jhkimnew, line 36: please add OnDenstory as well.

jhkimnew (approved):

Issue #391, opened 2019-06-19
UI issues with small window size


jhkimnew:
When we use small window size, manage.iis.net site support toggle Sidebar but in WAC mode, there is no way to show the side bar.
Another problem is the title string is truncated. 

![image](https://user-images.githubusercontent.com/20650163/59729647-288eb680-91f4-11e9-95dc-7881a71f21e7.png)

PR #399, opened 2019-06-21
Fixed UI issues

jhkimnew:
1. Fixed UI truncation issue of https://github.com/microsoft/IIS.WebManager/issues/391.
   When the navigation bar is hidden, we need to add extra padding to avoid the truncation issue.
2. Fix the accessibility issues regarding search button. The search button for Certificates feature is also refactored to fix the UI layout issue which was happened after fixing other accessibility bug of the unmatched element order.

### DIFF
--- a/src/app/common/vtabs.component.ts
+++ b/src/app/common/vtabs.component.ts
@@ -82,7 +82,7 @@ li:focus {
     font-weight: bold;
     margin-left: 1em;
     margin-top: 0.5em;
-    margin-bottom: 0;
+    margin-bottom: 0.5em;
     padding-bottom: 0;
 }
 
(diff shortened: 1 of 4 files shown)


src/app/common/vtabs.component.ts
```diff
@@ -53,7 +54,7 @@ import { Heading } from 'header/feature-header.component';
                 </ng-container>
             </ul>
         </div>
-        <div class="content sme-focus-zone">
+        <div class="content sme-focus-zone" [class.hidden-nav]="!IsActive">
```
jhkimnew: nav-hidden

shhsu (approved):

PR #403, opened 2019-06-24
Breadcrumb refactoring

shhsu:


src/app/common/feature-vtabs.component.ts
```diff
@@ -57,7 +55,7 @@ export class GlobalModuleReference {
     // NOTE: when [routerLink] is used, Angular automatically set tabindex="0". In order to avoid that behavior, we decided to add tabindex="-1".
     template: `
 <div class="sidebar crumb" [class.nav]="IsActive">
```
jhkimnew: please rename this to isActive from IsActive, too.

src/app/header/titles.component.ts
```diff
@@ -1,17 +1,17 @@
...
 
 @Component({
     selector: `titles`,
     template: `
 <div class="titles">
     <ul class="breadcrumbs">
         <li *ngFor="let crumb of breadcrumbs; index as i">
-            <span [ngClass]="{root: !i}" *ngIf="!(crumb.routerLink)">{{crumb.label}}</span>
-            <a *ngIf="crumb.routerLink" [routerLink]="crumb.routerLink">{{crumb.label}}</a>
+            <span [ngClass]="{root: !i}" [class.color-active]="canNavigate(crumb)" (click)="navigate(crumb)">{{crumb.label}}</span>
```
jhkimnew: I think you should add 'tabindex="0"' to make this focusable. Otherwise users can't navigate with keyboard.

jhkimnew: And add accessibility-focusable class as well to set the style when it is focused.

src/app/common/feature-vtabs.component.ts
shhsu: updated

jhkimnew (approved):

Issue #186, opened 2018-08-30
[Accessibility] A11y_Manage.IIS.Net_Local IIS-Server_Sensory Characteristics: Action buttons like "save", "delete", "cancel", "edit", "refresh" etc. has been provided by just using sensory icons and appropriate text label has not been provided to it.


accessibility18:
**Environment Details:**
URL: https://manage.iis.net/
Browser: Microsoft Edge 42.17134.1.0, Google Chrome Version 68.0.3440.106

**MAS ID Violated**: MAS1.3.3

**Repro Steps:**
1. Download and install "Microsoft IIS Administration" application.
2. Enable Internet information Services.
3. Open "Microsoft IIS" application using the URL: https://manage.iis.net/
4. Tab till "Local IIS" page tab and Press Enter.
5. Verify that only sensory elements are not used for any of the UI element.

**Actual Result**:
Following UI controls has been provided by just using sensory icons and appropriate text label has not been provided to it: "Show/Hide sidebar", "Edit", "Delete", "More", "Save", "Cancel", "Refresh", "Upload", "New Root", "Go Up" buttons etc.

**Expected Result:**
Instead of using just the sensory icons for buttons, there should be appropriate text label and a rectangular border around the buttons.

**User Impact:**
Users who are not aware of these icons may not understand the function of the controls properly.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={1823da5a-e8fb-4e63-b881-7eccb312c358}

**Attachment:**
![bug 186_snesory icons are used for the ui components-1](https://user-images.githubusercontent.com/38521133/44843955-370ce100-ac67-11e8-9961-13e21e120c07.png)
![bug 186_snesory icons are used for the ui components-2](https://user-images.githubusercontent.com/38521133/44843956-37a57780-ac67-11e8-80e1-930aa80f9f6d.png)
![bug 186_snesory icons are used for the ui components-3](https://user-images.githubusercontent.com/38521133/44843958-37a57780-ac67-11e8-8fe4-f9ed6da6e5e8.png)

jhkimnew:
Fixed

mstechie:
GitHubTags:#A11y_Manage.IIS.Net;

PR #415, opened 2019-06-27
Fixed missing area-hidden for toolbar and tabindex issue of tabs

jhkimnew:
1. Fixing the issue for https://github.com/microsoft/IIS.WebManager/issues/186
2. Fixing the tabindex issue. The hidden tabMenu button should not be focused when it becomes hidden.

### DIFF
--- a/src/app/common/tabs.component.ts
+++ b/src/app/common/tabs.component.ts
@@ -180,12 +180,12 @@ import { SectionHelper } from './section.helper';
                             <span class="border-active">{{tabs[_selectedIndex].name}}</span>
                         </li>
                     </ul>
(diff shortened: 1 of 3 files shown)


src/app/common/tabs.component.ts
```diff
@@ -287,6 +292,16 @@ export class TabsComponent implements OnDestroy, AfterViewInit {
         this._sectionHelper.selectSection(this.tabs[index].name);
     }
 
+    isTabMenuHidden() {
```
shhsu, line 295: shrink the logic?

shhsu (approved):

Issue #219, opened 2018-10-17
Create a new IIS.WebManager release


bariscaglar:

shhsu:
@shirhatti  has already deployed 2.2.3

bariscaglar:
Reopening as discussed offline.

shhsu:
TODO: fix settings/servers page is broken due to missing heading.

PR #430, opened 2019-07-10
Fix site mode

shhsu:


src/app/main/app-routing.module.ts
```diff
@@ -2,22 +2,18 @@ import { NgModule } from "@angular/core";
...
 import { WebServerRoute } from "./settings";
 
 @NgModule({
   imports: [
     RouterModule.forRoot([
-        // site only routes
-        { path: "", component: HomeComponent },
-
         // common routes
+        { path: "", redirectTo: "/webserver", pathMatch: "full" },
```
jhkimnew, line 12: Make sure web server/services still works after removing homecomponent.

jhkimnew (approved):

src/app/main/app-routing.module.ts
shhsu: This PR is to unblock fixing some accessibility issue. We will fix this page if we decide to proceed with
https://github.com/microsoft/IIS.WebManager/issues/219

PR #441, opened 2019-07-18
Make grid item name columns clickable

shhsu:


shhsu:
Previous revision file selection. It will be added in next revision

src/app/files/file-list-item.ts
```diff
@@ -4,15 +4,17 @@ import { Humanizer } from 'common/primitives';
...
-import { ActivatedRoute } from '@angular/router';
 
 @Component({
     selector: 'file',
     template: `
         <div *ngIf="model" class="grid-item row" [class.background-editing]="_editing && !_location" [class.background-selected]="_editing && _location" (keyup.f2)="onRename($event)" tabindex="-1">
             <div class="col-xs-9 col-sm-5 col-lg-4 fi" [ngClass]="[model.type, model.extension, (isRoot ? 'location' : '')]">
                 <div *ngIf="!_editing || _location">
-                    <a tabIndex="0" class="color-normal hover-color-active" nofocus><i></i>{{model.alias || model.name}}</a>
+                    <a tabIndex="0"
```
jhkimnew, line 14: Consider the accessibility of showing underline when we hover or set input-focus.

jhkimnew (approved):

src/app/files/file-list-item.ts
shhsu: updated

Issue #386, opened 2019-06-12
New File Root dialog does not work


shhsu:
1. the dialog box closes itself frequently
2. It doesn't seem to modifying the config file at all

shhsu:
Confirmed with @jhkimnew 
It was a UI glitch causing config file to not be updated

shhsu:
fixed by https://github.com/microsoft/IIS.WebManager/pull/444

Issue #434, opened 2019-07-17
Don’t close modal when you click outside the modal


shirhatti:
Modals should only be closed on an explicit gesture like clicking a cancel button

shhsu:
https://github.com/microsoft/IIS.WebManager/pull/444

PR #444, opened 2019-07-22
Selectors should not always close when somewhere else is clicked

shhsu:
https://github.com/microsoft/IIS.WebManager/issues/434

### DIFF
--- a/src/app/files/edit-location.component.ts
+++ b/src/app/files/edit-location.component.ts
@@ -6,7 +6,7 @@ import { Location } from './location';
 @Component({
     selector: 'edit-location',
     template: `
-        <div class="sme-focus-zone">
+        <div>
             <fieldset class="name">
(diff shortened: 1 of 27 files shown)


src/app/files/edit-location.component.ts
```diff
@@ -6,7 +6,7 @@ import { Location } from './location';
 @Component({
     selector: 'edit-location',
     template: `
-        <div>
+        <div class="sme-focus-zone">
```
jhkimnew: We don't need to use sme-focus-zone and arrow key can be used instead.

shhsu: removed, tested, seem to still work

shhsu:
This PR will also fix https://github.com/microsoft/IIS.WebManager/issues/386 as well

jhkimnew (approved):

PR #449, opened 2019-07-24
Better handle lists loading/empty

shhsu:


jhkimnew (approved):
Please update the virtual directory feature as well.

PR #466, opened 2019-07-26
Authentication page error fix

shhsu:


src/app/webserver/authentication/authentication.service.ts
```diff
@@ -140,70 +134,34 @@ export class AuthenticationService {
...
-                this.anonymousError = e;
-                throw e;
-            })
+        return this.loadSafe<AnonymousAuthentication>("anonymous", this._anonAuth);
     }
 
     private loadBasic(): Promise<BasicAuthentication> {
-        return this._http.get(this._settings.getValue()._links.basic.href.replace("/api", ""))
-            .then(basic => {
-                this._basicStatus = Status.Started;
```
jhkimnew, line 158: Missed the below lines while refactoring into the loadSafe() function.
this._basicStatus = Status.Started; 
this._digestStatus = Status.Started
this._windowsStatus = Status.Started;

src/app/webserver/authentication/anon-auth.component.ts
```diff
@@ -26,16 +27,26 @@ import { AuthenticationService } from './authentication.service';
...
+        private _notificationService: NotificationService,
+    ) {
+    }
+
+    public ngOnInit() {
+        this._subscriptions.push(this._service.anonAuth.subscribe(
+            auth => this.setFeature(auth),
+            e => {
+                this._error = e;
+                this._notificationService.apiError(e);
```
jhkimnew, line 48: One issue here. NotificationService message is disappeared after timeout and users will lost the error message. Can we create a new notificationservice which does not have timeout?

src/app/notification/notification.service.ts
```diff
@@ -122,6 +122,12 @@ export class NotificationService {
             return;
         }
 
+        // Do not show programming error. They will be logged on the console window
+        if (error instanceof TypeError) {
+            this.warn("An unknown error has occurred while processing your request.");
```
jhkimnew: Instead of not showing anything from the original error message/detail, how about showing some part of it so that users can find what error happened.

src/app/webserver/authentication/anon-auth.component.ts
shhsu: timeout will be removed in the next PR

src/app/notification/notification.service.ts
```diff
@@ -122,6 +122,13 @@ export class NotificationService {
             return;
         }
 
+        // Do not show programming error. They will be logged on the console window
+        if (error instanceof TypeError) {
+            // TODO: create a wiki page and point to the troubleshoot wiki instead
```
jhkimnew, line 127: Did you create a work item of "create a wiki page and point to the troubleshoot wiki"?

jhkimnew:
Please put some description on this PR in order to explain the purpose of this PR. 
Don't forget to make issues to update other features which have the same problem.

jhkimnew (approved):
@shhsu Please note that I added some additional comments to remind you to make new follow-up issues because this fix is only for Authentication feature page, not for all, which has the same problem.

shhsu:
https://github.com/microsoft/IIS.WebManager/issues/467

Issue #464, opened 2019-07-25
Upload/Download, New/Edit File does not work because the request serialization is broken for the file content.


jhkimnew:
@shhsu Currently Upload/Download and New/Edit File does not work if the file is saved using UTF-8 with BOM. If we can't fix this issue, I think we should disable the Upload/Download and New File buttons and then remove the link for the physical file in order to disable the Edit File feature.

shhsu:
For context, because WAC mode uses powershell to proxy the requests and encode http requests as json, it messes up any file edit/download/upload for any files that are not utf-8

shhsu:
The proper solution is to bypass IIS Admimistration API and use WAC's FileTransferService

PR #469, opened 2019-07-29
Jhkim/issue 464

jhkimnew:
https://github.com/microsoft/IIS.WebManager/issues/464

1.	IIS.WebManager uses JSON.stringify() to serialize the JSON object of request information. For the File System API Put scenario, we used ArrayBuffer to send request body and that caused the issue to the JSON.stringify. 
It simply does not support the data type and body data was lost because of the issue. So, as a work-around, I have updated the data type from ArrayBuffer to Uint8Array. 
2.	For the File System API Get scenario, we serialized the respond body with byte[] array and that caused the data loss while deserializing if the file content is not pure UTF8 data. As a work-around, we should read the response body in string, not in byte[], to support various encoding such as UTF16 and UTF8, even UTF8 BOM characters. 
3.	In uploading scenario, there is a bug while handling the file attribute of lastUpdatedDate. We incorrectly used the number value instead of the supported ISO Date format string value. I have converted the number value with ISO Date format string value.

However, there is one remaining issue regarding downloading. The download seems not to be possible in the current design. We send the downlevel request to the backend server and the downlevel is started from the backend sever, not from the user computer. 
Because there is no easy fix for the downloading issue, I suggest that we should disable the downloading button for this time and consider to fix in the next milestone.


src/app/resources/scripts/iis_scripts/Invoke-LocalHttp.ps1
```diff
@@ -67,8 +67,19 @@ $uri = $uriBuilder.ToString()
...
 try {
     $httpMsg = New-Object System.Net.Http.HttpRequestMessage -ArgumentList $httpMethod, $uri
-    if ($reqObj._body) {
-        $requestContent = New-Object System.Net.Http.StringContent ([string] $reqObj._body)
+    if ($reqObj._body -or $reqObj._body_Uint8Array) {
+        if ($reqObj._body_Uint8Array) {
+          #
+          # convert $reqObj._body_Uint8Array to ByteArray(byte[])
+          #
+          $byteArray = new-object byte[] $reqObj._body_Uint8Array_Length
```
shhsu: `New-Object` for consistency

src/app/resources/scripts/iis_scripts/Invoke-LocalHttp.ps1
```diff
@@ -67,8 +67,19 @@ $uri = $uriBuilder.ToString()
...
+          for ($i = 0; $i -lt $reqObj._body_Uint8Array_Length; $i++) {
+            $byteArray[$i] = $reqObj._body_Uint8Array.$i;
```
shhsu: $reqObj._body_Uint8Array[$i] for consistency

shhsu (commented):
Did we need to also encode the body on the way back?
Have you tried the download functionalities to download a binary file

src/app/resources/scripts/iis_scripts/Invoke-LocalHttp.ps1
jhkimnew: Good catch. Fixed.

src/app/resources/scripts/iis_scripts/Invoke-LocalHttp.ps1
```diff
@@ -135,4 +151,4 @@ if ($result) {
     $result
 } else {
     throw "Unexpected error occured"
-}
+}
```
shhsu: nit: newline

src/app/runtime/wac/services/powershell-service.ts
```diff
@@ -132,4 +142,4 @@ export class PowershellService {
       }),
     );
   }
-}
+}
```
shhsu: nit: newline

shhsu (approved):
add back newlines and remove download button

src/app/resources/scripts/iis_scripts/Invoke-LocalHttp.ps1
jhkimnew: fixed

PR #489, opened 2019-08-06
Add CredSSP for network installation

shhsu:


jhkimnew:
@shhsu Would you adjust the label of the installation UI with this bug fix together if you don't mind. The first label should like Microsoft IISAdministration API and the next two labels should have "(Optional)". 

image

src/app/runtime/wac/components/install.component.ts
```diff
@@ -167,20 +169,18 @@ export class InstallComponent implements OnInit {
                 return `${fieldName} cannot be empty`;
             }
         }
-
-        // NOTE: This code is injected here to prevent user using shared drive because it would be considered double hop with WinRM and wouldn't work
-        // This needs to be revisited when Windows Admin Center completed their work on CredSSP
         if (location && location.startsWith(`\\\\`)) {
```
jhkimnew, line 172: Some minor issue. If we don't trim() the given path string value, "  \\test...", which is an example valid UNC path, will be considered as an invalid value. How about trim() the value before using it?

jhkimnew (approved):
@shhsu I found an minor issue. Please consider to fix the issue if you agree.

src/app/runtime/wac/components/install.component.ts
shhsu: fixed

PR #493, opened 2019-09-11
Add preview badge

shhsu:


src/app/common/vtabs.component.ts
```diff
@@ -12,10 +13,10 @@ import {
     QueryList,
     ViewChildren,
 } from '@angular/core';
+import { FormsModule } from '@angular/forms';
 import { ActivatedRoute, Router, RouterModule } from '@angular/router';
-import { CommonModule, Location } from '@angular/common';
+import { BadgeModule } from '@microsoft/windows-admin-center-sdk/angular';
```
jhkimnew, line 18: this is not required.

src/app/common/vtabs.component.ts
```diff
@@ -364,6 +437,8 @@ export const TABS: any[] = [
         CommonModule,
         DynamicModule,
         TitlesModule,
+        BadgeModule,
```
jhkimnew, line 440: BadgeModule is not required

jhkimnew (approved):
Reminding...  
Put some commit comment that we should revert this commit later to remove the badge.

src/app/common/vtabs.component.ts
shhsu: fixed

src/app/common/vtabs.component.ts
shhsu: fixed

PR #520, opened 2022-08-29
Make the project build for WAC

yaqiyang:
The project should be built for WAC when --env=wac is used with build.ps1. But it doesn't work. I don't see any code in the build process that handles --env=wac. After inspecting the code, I renamed the following files 
environment.ts is from environment.wac.prod.ts
app.addon.ts is from app.addon.wac.ts. The original file is saved app.addon.site.ts
app-routing.module.ts is from app-routing.module.wac.ts. The original file is saved as app-routing.module.site.ts
bootstrap.ts is from bootstrap.wac.ts. The original file is saved as bootstrap.site.ts.

In the future, if we want the project to work as standalone web application, we need to fix the build process to copy the files according to the --env switch

### DIFF
--- a/src/app/environments/environment.ts
+++ b/src/app/environments/environment.ts
@@ -1,2 +1,2 @@
-export const IsProduction = false;
-export const IsWAC = false;
+export const IsProduction = true;
+export const IsWAC = true;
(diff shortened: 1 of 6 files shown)


yaqiyang:
No need to change these files. Need to call guild build with -c=wac or --configuration=wac

PR #521, opened 2022-08-29
Do not prompt for version specific IIS Administration

yaqiyang:
The user will automatically get IIS Adminstration from http://go.microsoft.com/fwlink/?LinkId=829373 or download from https://github.com/microsoft/IIS.Administration/releases. We don't indicate a specific version here


yaqiyang:
The new UI

New-IIS-Manager

yaqiyang:
The original UI
image

John-Hart (approved):

